Issue #61, opened 2014-07-27
Node kits can only be constructed from parts known at compile time


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/61) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

----------------------------------------

In Coin the macro SO_KIT_ADD_CATALOG_ENTRY() directly calls getClassTypeId() of the part. This restricts node kits to be constructed only from parts for which include files are available. Using nodes registered with the scene object database by other extensions is currently unsupported. To improve compatibility with other Open Inventor implementations it would be better to look up the part type via SoType::fromName() as already done in the SO_KIT_CONSTRUCTOR() macro.

PR #194, opened 2019-12-23
Use SoType::fromName() in SO_KIT_ADD_CATALOG_ENTRY()

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/11) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

Solution for issue #61. Look up type in SO_KIT_ADD_CATALOG_ENTRY() via SoType class to avoid compile time dependencies to all parts. Tested change on Win7 and OS X (use of SoTrackballManip).


VolkerEnderlein:

This looks OK. I will also test it in my own viewer application (running on Win8.1) - I can try all the manipulators very easily.

Any observations on impact on execution speed?

VolkerEnderlein:

I didn’t observe any performance impact in normal use cases. But to be sure I wrote a simple test program that creates 100000 instances of SoShapeKit (15 catalog entries each). Without the change this took 8.19 seconds on my laptop. With the type lookup via SoType::fromName() from the pull request the same loop took 9.16 seconds, which means it adds about 10%. Note though that the node configuration by the application is where typically most of the time is spent. For example, initializing those 100000 shapes with orientation, color and position as in the “Random Discs” example (https://github.com/TheHubbit/PyInventor/wiki/PyInventor-Examples) takes about 30 seconds. In performance critical situations (several thousands of scene objects) node kits might not the best choice to begin with (creating 100000 SoSphere instances for example is done in less than half a second).

Furthermore I checked the header files of the original SGI and also VSG inventor libraries. Both are looking up the type via SoType class instead of calling getClassTypeId() directly. I believe the performance penalty is not relevant for normal use cases and worth accepting given the compatibility gain to the original Inventor library.

In case this is not acceptable an alternative proposal would be to add another macro called SO_KIT_ADD_CATALOG_ENTRY_EXT() that does the type lookup for external nodes to be added to a node kit.

VolkerEnderlein:

For what it's worth, I like the looks of it, and I appreciate your contribution and thoroughness. I think this is acceptable.

VolkerEnderlein:

Thanks for that detailed work. I will accept this.

Issue #68, opened 2014-08-27
SbString constructor for wide character pointers


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/68) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

----------------------------------------

Starting with version 4 Coin3D uses UTF8 encoded strings. However, there is no method exposed by the SbString class that encodes a wide character string to UTF8. Also the included utility function cc_string_utf8_encode() only converts single characters and does not handle platform specific encoding of wide character strings (Linux: utf32, Windows: utf16).

In my opinion it would be very useful if the SbString class accepts also wide character strings, taking care of the conversion to UTF8. Please comment on this proposal. Would it be safe to assume that all targeted compilers support the wchar_t type now? I could implement this functionality if there are no concerns.

PR #198, opened 2019-12-23
Added SbString constructor for wide character strings.

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/15) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

This pull request addresses issue #68. A new constructor has been added to the SbString class that accepts wide character strings.


VolkerEnderlein:

Thomas,

Once again, many thanks for your coding. I have reviewed the pull request code. My only comment is that line 137 (updated numbering) is redundant due to the earlier if statement.

As for further ramifications - I was wondering, for example, if functions such as 'cc_string_length' and 'cc_string_is' always work correctly even for all possible wide strings. I haven't actually tried this yet so I might need somewhat longer to investigate.

Roy

VolkerEnderlein:

Roy,

Thank you for reviewing! I really appreciate it. Good catch regarding the redundant if statement. It was also very misleading as in this case the internal buffer should never been freed. I updated the code now.

The string functions will work on UTF8 encoded strings as they would do normally. But the length will return the number of characters needed for the encoded string, not the length of the wide character string (so it could be longer). I also haven't dealt much with UTF8 encoding before and found this article very useful: http://www.nubaria.com/en/blog/?p=289

Thomas

VolkerEnderlein:

Thomas,


Thanks again for the code update. 


That was an excellent article you linked to.


Roy

VolkerEnderlein:

While testing the string functions I came across another issue in cc_string_compare_text(). Instead of returning the result of strcmp() - as done by cc_string_compare_subtext() - it performs a logical and with the input arguments. I assume this was done to avoid calling strcmp() with null pointers. However, the return value is altered and thereby incorrect in case one of the arguments is null or the first string is smaller than the second one. I corrected this by checking each argument individually and treat them as an empty string in case of null pointers.

PR #186, opened 2019-12-23
Use CMake build system

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/3) by Christian Bühler (Bitbucket: [cbuehler](https://bitbucket.org/cbuehler), GitHub: [cbuehler](https://github.com/cbuehler)).**

----------------------------------------

Tested with Windows 7 / MSVC 2013 and Arch Linux / gcc 4.8.2 (together with pull request #186).
Could someone please test on Mac (@fspindler?)?

Does not include all the functionality of Autotools yet, but the .dll/.so gets compiled.

PR #200, opened 2019-12-23
add CMake build system

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/17) by Christian Bühler (Bitbucket: [cbuehler](https://bitbucket.org/cbuehler), GitHub: [cbuehler](https://github.com/cbuehler)).**

this replaces pull request #186


VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Nice try, but is it necessary to split coin in 41 libraries? Instead of add_library in each folder it would be possible to add the files to a variable and create one library in the end.

Also, your include/Inventor/system/CMakeLists.txt is wrong. Instead of

``` if(WINDOWS) ```

you have to use

``` if(WIN32) ```

the current code is not working on Windows with MSVC. And I would add

``` set(CMAKE_DEBUG_POSTFIX "d") ```

to create different libraries for debugging and release.

The install part needs to be updated. Because the include files are not installed.

An other improvement would be a switch to change between shared and dynamic library.

VolkerEnderlein:

Christian,

Sorry to be a pain, but can you modify this request, please, so that is goes to the newly created CMake branch rather than the default? It seems that I can't change it because it is your pull request. Note that the branches have been synced.

We have decided that we need to investigate this more thoroughly and your code is a good starting point.

Roy

VolkerEnderlein:

The complaint may be how this generates the VS solution, where every directory becomes a separate project.

Issue #83, opened 2014-11-29
SoShaderStateMatrixParameter does not support GL shaders


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/83) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

----------------------------------------

The class SoShaderStateMatrixParameter is limited to Cg and currently doesn't support GLSL shaders. This can be fixed as noted in a comment by *mortene* a while back. Furthermore the class documentation for shader parameters is incomplete.

Issue #44, opened 2013-11-06
Make the documentation available again


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/44) by Alexander_F (Bitbucket: [Alexander_F](https://bitbucket.org/Alexander_F), ).**

----------------------------------------

Make a working link to the Coin documentation available, since http://doc.coin3d.org/xx as stated in the Wiki doesn't work anymore.

VolkerEnderlein:
As the docs are up and running again on [Bitbucket](https://coin3d.bitbucket.io/) and [GitHub](https://coin3d.github.io/) I am closing this issue. Feel free to reopen it if the issue has not properly addressed. Otherwise please open a new issue on our GitHub [issue tracker](https://github.com/coin3d/coin/issues).

PR #205, opened 2019-12-23
Updated shader parameter documentation and fix for issue #83

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/22) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

This pull request contains new and updated class documentation, especially of the [shader parameter](http://coin3d.bitbucket.org/Coin/classSoShaderParameter.html) classes. It also enhances SoShaderStateMatrixParameter with support for GLSL (before only Cg was supported, see issue #83). Last but not least I updated the doxyfile under msvc10 based on *docs/coin.doxygen.in* because the doc creation failed otherwise.


VolkerEnderlein:

Thomas,

Great job once again. I have been reviewing your codes below and have two questions.

I notice that in SoShaderParameter.cpp you have a number of default: assert(FALSE); statements in your 'case' clauses. Are you happy that none of these require full error reporting?

I have been using doxygen 1.8.8, which (currently) is the latest version, for my work on issue #44. I notice that your code was for 1.8.7. I think we should all use the same version. Do you have any preferences?

Roy

VolkerEnderlein:

Roy,

Thank you for praising my work and reviewing the changes. I really appreciate it!

The GLSL/ARB conditional branch is based on the code branch for Cg. I did copy that 1:1 including the assert statements trying to keep the style. Searching a bit through Coin source files I also found that this style is being used quite often: 
```
#!c++
assert(0 && "unknown alignment");
```
I now changed the assert statements in SoShaderParameter.cpp to this style as it is more descriptive. However, I wouldn't report unknown enums via SoError classes. This is a critical internal error and not something that could have been caused by the user of the library (invalid input or similar). Stopping the execution with an assert seems to be appropriate to me.

Regarding the doxygen file under msvc10. I based it on the original [coin.doxygen.in](https://bitbucket.org/Coin3D/coin/src/f7257e49e9e5fc83df96cae27540735ad1f39957/docs/coin.doxygen.in?at=default) file that is used when building with the make system. I'm using version 1.8.8 of doxygen too and the configuration works fine with that. It just seems that you generated the doxygen file with version 1.8.7 and switched to 1.8.8 later. Actually, I now also played a bit with the generation of the documentation and would like to make some more changes. But I will comment on that further in the issue #44 thread.

Thomas

VolkerEnderlein:

As you may know, Doxygen configuration files can be updated to a newer version with the -u flag.

Issue #71, opened 2014-09-17
Add --with-boost configuration flag


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/71) by Nathan Litke (Bitbucket: [njlitke](https://bitbucket.org/njlitke), GitHub: [njlitke](https://github.com/njlitke)).**

----------------------------------------

I'm building the current top-of-tree (d13a8803c4bd) on OS X 10.9.4.  I've installed boost through MacPorts in its default location, /opt/local/include.  The only way I've found to configure the build to find boost by setting CXXFLAGS:

../configure CXXFLAGS="-I/opt/local/include"

As with other external dependencies, it would be nice to have a flag to specify the path to boost, e.g.,

../configure --with-boost=/opt/local/include

Thanks,
Nathan

PR #209, opened 2019-12-23
Fix build issues about headers and casting.

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/26) by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**


VolkerEnderlein:

Hi Alexey,

Thanks for splitting the pull request into two.

I have had a little play with your first code change below, i.e. line 59 of spidermonkey.h. I am still using Visual Studio 2008 (Standard Version, at home). If I substitute the 'long' type for 'intmax_t' my compiler fails, not surprisingly. Can we include a definition for this please if it is not already defined? This would be primarily in 'Include\Inventor\system\inttypes.h.in'. Obviously the individual compiler dependent autogenerated files in the build directories would also need updating.

Roy

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

I think we can use "int64_t" instead "intmax_t"

VolkerEnderlein:

Alexey,

Now it becomes a bit easier to digest. I imported your changes to test them with Visual Studio 2010 compiler. Here I ran into the same problem as Roy with VS2008. With a little bit more time to review, I also have some more remarks and questions.

*spidermonkey.h:*
The intmax_t type has been added recently and is not supported by VS2010. However, it is defined in the include file boost/cstdint.hpp. Maybe this could resolve the issue while staying compatible to different compilers and platforms? I haven't used scripting with coin yet and can't validate this though.

*dict.h / glyph.cpp / glyph?d.cpp:*
Your change from the boost includes (cassert and cstdlib) worked fine in my test. But I believe boost was used here for improved compatibility with a variety of platforms. Is the build problem that you're experiencing related to the fact that the included boost is getting a little old? Could this be related to Issue #71 (can be solved with newer boost version)?

*freetype:*
Even though I haven't built with freetype, switching to the freetype2 include directory makes sense. I will try to test that on OSX at some point where I have those includes available.

*SoSTLFileKit.cpp:*
Hopefully I understood this better now. The issue is caused because coin stores integer numbers in void pointer variables (user data) and some compilers don't allow the simple cast of a 64 bit pointer to a 32 bit integer. Despite being kind of ugly it seems to be OK in this case as the stored indices are always 32 bit integers. A difference between two pointers also results in a number (not pointer), which can then be casted down. Maybe we can change the code to something like this to avoid using intmax_t type and forcing 64 bit operations on 32 bit builds.

```
#!c++
  // convert a pointer to a numerical value by "computing" the difference to a null pointer
  #define COIN_VOIDPTR_TO_INT(ptr) (((const char*) ptr) - ((const char*) 0))

  long v1idx = PRIVATE(this)->points->findPoint(v1), v1new = (v1idx == -1);
  long v2idx = PRIVATE(this)->points->findPoint(v2), v2new = (v2idx == -1);
  long v3idx = PRIVATE(this)->points->findPoint(v3), v3new = (v3idx == -1);
  if (!v1new) { v1idx = (long) COIN_VOIDPTR_TO_INT(PRIVATE(this)->points->getUserData(v1idx)); }
  if (!v2new) { v2idx = (long) COIN_VOIDPTR_TO_INT(PRIVATE(this)->points->getUserData(v2idx)); }
  if (!v3new) { v3idx = (long) COIN_VOIDPTR_TO_INT(PRIVATE(this)->points->getUserData(v3idx)); }
  long nidx = PRIVATE(this)->normals->findPoint(n);
  if (nidx != -1) { nidx = (long) COIN_VOIDPTR_TO_INT(PRIVATE(this)->normals->getUserData(nidx)); }
```

*thread.cpp / thread_win32.icc:*
Same as with SoSTLFileKit. The TLS value is used to store the thread id. I would convert it after computing a difference with a null pointer. Could you try if it works in your configuration?

*JS_VRMLClasses.cpp:*
Looks good to me but I can't test it (don't know how to). But it does compile in my environment.

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

*spidermonkey.h:*
As I told earlier we can try to use "int64_t". It present in 'Include\Inventor\system\inttypes.h.in'

*dict.h / glyph.cpp / glyph?d.cpp:*
Headers "cassert" and "cstdlib" is included not from boost. This is standard headers. Read:
[http://stackoverflow.com/questions/2900785/whats-the-difference-between-cstdlib-and-stdlib-h](http://stackoverflow.com/questions/2900785/whats-the-difference-between-cstdlib-and-stdlib-h)

*SoSTLFileKit.cpp:*
I think we all need read 'Include\Inventor\system\inttypes.h.in' header. There is "COIN_INTPTR_T" type. This type have always correct bitsize for pointers for platform.

*thread.cpp / thread_win32.icc:*
I think there is all ok. Just need replace "intptr_t" with "COIN_INTPTR_T"

VolkerEnderlein:

Alexey,

Thanks for your reply and proposals! Still I have a few more questions:

*spidermonkey.h*: Did you mean "typedef intptr_t jsword;"? It should still be a 32 bit type for those systems, right? I gave that a try and it compiles fine for me on Windows with VS2010.

*dict.h / glyph.cpp / glyph?d.cpp*: The link was helpful. I also found answer 24 in [this](http://stackoverflow.com/questions/10625716/in-what-cases-we-need-to-include-cassert) discussion interesting. However, it seems to hint that some versions of MinGW g++ are quite non-standard with those includes. I searched in the Coin code and it seems that the use of "using namespace ..." is avoided intentionally. The c-includes use the std namespace. Therefore "using std::free;" and such should work. What error did you run into with MinGW g++? I'm still trying to understand the root cause.

*SoSTLFileKit.cpp*: Using a 64 bit number as in the pull request below triggers a lot of warning later on because the indices are used as function arguments where 32 bit integers are expected. I would therefore leave the variables as is and only change the casting as shown here (similar to your change in the thread classes):
```
#!c++
  // find existing indexes if any
  long v1idx = PRIVATE(this)->points->findPoint(v1), v1new = (v1idx == -1);
  long v2idx = PRIVATE(this)->points->findPoint(v2), v2new = (v2idx == -1);
  long v3idx = PRIVATE(this)->points->findPoint(v3), v3new = (v3idx == -1);
  if (!v1new) { v1idx = (long) reinterpret_cast<intptr_t>(PRIVATE(this)->points->getUserData(v1idx)); }
  if (!v2new) { v2idx = (long) reinterpret_cast<intptr_t>(PRIVATE(this)->points->getUserData(v2idx)); }
  if (!v3new) { v3idx = (long) reinterpret_cast<intptr_t>(PRIVATE(this)->points->getUserData(v3idx)); }
  long nidx = PRIVATE(this)->normals->findPoint(n);
  if (nidx != -1) { nidx = (long) reinterpret_cast<intptr_t>(PRIVATE(this)->normals->getUserData(nidx)); }
```

Does this compile in your configuration too? Like this the file compiles without warnings on my system.

*thread*: I'm sorry, my mistake. The change from your pull request works for me (intmax_t is not defined, but intptr_t is!).

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

Any plans to merge it?

VolkerEnderlein:

Alexey,

This file is showing a conflict for me. Are you seeing it too? If so, what do you think the resolution should be?

Roy

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

I think we need use safe casting as in my patch.

VolkerEnderlein:

Am I correct in thinking that this is the second of the two alternatives?

Roy

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

Yes, you need use
````return static_cast<unsigned long>(reinterpret_cast<intptr_t>(val))````

**Comment by VolkerEnderlein** [comment]
**Original comment by Edward Smith-Rowland (Bitbucket: [Ed Smith-Rowland](https://bitbucket.org/Ed Smith-Rowland), ).**

I was doing something like this 
My approach was to substitute

```
#!c++

    long -> intptr_t
```

and

```
#!c++

    unsigned long -> uintptr_t
```
I'll post my patch when I get to my other computer.

These integer types are designed for pointer to integer conversion.
These should be portable as well - they are standard (forgot which).
I've only tested it on mingw-w64 both i686 (32-bit) and x86_64 (64-bit).
I'll try it on linux and msvc unless someone beats me to it.

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

Problems not resolved for this years. I delete repository and will leave patches in my build scripts

VolkerEnderlein:

As I see it, there were still a few open questions, or did I misread? Now that the repository is deleted, I can’t see the patch anymore.

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

If anyone still interesting all patches arrived to [https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-coin3d-hg](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-coin3d-hg)

VolkerEnderlein:

Great, thanks a lot.

VolkerEnderlein:

Hi Alexey, I am planning to merge your patches to the code base finally. In the meantime lots of changes have been made to the Coin repository, especially the build system scripts.

I was able to get it to build successfully but a few questions still remain. 

What is the supposed output name of the DLL under MSYS2/mingw-w64 ? Is it libCoin4.dll (major version included), libCoin-80.dll (SOVERSION included), or simply libCoin.dll? From the Coin 3.1.3 autotools script I got the name libCoin-60.dll (SOVERSION included) so for Coin 4.0.0 libCoin-80.dll seems quite likely.

And what are the proposed names for import library (libCoin.dll.a, I suppose) and static library (libCoin.lib, I suppose)?

VolkerEnderlein:
**Original comment by Alexey Pavlov (Bitbucket: [Alexey Pavlov](https://bitbucket.org/Alexey Pavlov), ).**

Currently I use CMAKE build system, so dll named libCoin.dll

For all MINGW platforms standard suffixes are:

.dll.a - for import library

.a - for static library

VolkerEnderlein:

Back on track for this task. Thank you for clarification, Alexey. Somehow I mixed up the static library suffix when writing that e-mail.

AFAIK, only the names of the import library or the static library are relevant to the linker. As we are currently supporting both, the autotools and the CMake script, is it OK with you to follow the naming scheme of the old autotools script and name that DLL libCoin-80.dll? The reason I ask is, that as an end user I would be very surprised to see different build system generators output different artifact names on the same project.

Issue #89, opened 2015-01-15
Support DISABLE policy in SoSelection


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/89) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

----------------------------------------

The SoSelection node currently does not support disabling selection changes via the policy field. However, applications commonly need this and it is also supported by VSG Inventor for example (see [class reference](http://oivdoc94.vsg3d.com/APIS/RefManCpp/class_so_selection.html)). Therefore I think Coin should be enhanced to support the DISABLE policy as well.

PR #211, opened 2019-12-23
Support DISABLE policy in SoSelection node

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/28) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

Added DISABLE option to policy of SoSelection node. Resolves issue #89.


VolkerEnderlein:

Nice. Should we deselect any existing selections when policy is set to DISABLE?

VolkerEnderlein:

Bastiaan,

Thank you! I would prefer not to deselect all nodes on policy change. It is easy for the application to call deselectAll() on the selection node before or after changing the policy. However, sometimes you may just want to "freeze" the current selection. Also, right now invokeSelectionPolicy() is called by handleEvent() and not when the field changes. I think it would be a bit strange if the deselection is not happening right away but after the next mouse click.

VolkerEnderlein:

Thomas,

I see, good points.

You are right in the feature request, I have implemented disabling selection myself as well, but with a different API. I keep a counter that is increased with each call to disable(), and selection is enabled only after an equal number of calls to enable(). However I think it makes sense to follow VSG's API.

VolkerEnderlein:

Looks good to me.

PR #217, opened 2019-12-23
Updated CMake files

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/34) by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

* allow single project build
* allow multithreaded build (MSVC)
* create documentation
* install files, header and documentation
* PDB for debugging (MSVC)
* add RC for Windows DLL
* create headers from glsl and iv


VolkerEnderlein:

transporter,

Great work. One question at the moment. What is the motivation for adding files to a new "src/internal/data" directory? They were already in the "data" directory.

Roy

VolkerEnderlein:

Hi Roy,

The files in src/internal/data have been generated from the files in data. I guess it is similar to the data folders of the msvcXX build folders as the sh command is not available on all platforms.

VolkerEnderlein:

Thank you transponder for your work on the build system! Very nice! I too noticed the newly added files in src/iternal/data. Will all platforms use those files now instead of generating them from the iv and shader files? Maybe this could be done depending on whether the sh command has been found or not?

Another observation I have is related to the COIN_BUILD_SINGLE_LIB option. I'm not so thrilled about this option as it causes a lot of duplication in all the CMakeLists.txt files, which I feel could be a source of errors in the future. Even when using add_library() only one shared library is created. Furthermore the smaller input libs for the Coin library can be build in parallel (at least using MSVS solutions). To me it does not seem too bad. Last but not least I fear we would run into a command line length limitation of the compiler at some point when linking all object files at once (see [here](https://msdn.microsoft.com/en-us/library/610ecb4h.aspx)).

I approve of all the other changes you've made (version numbers, rename options, adding missing compiler definition for DLL and of course generating documentation). And if others agree with introducing the COIN_BUILD_SINGLE_LIB option - or you feel strongly about it - I wouldn't mind leaving this in (as long as it works ;-) ).

Thomas

VolkerEnderlein:

We should be able to generate these files from the .iv files on any platform, and there are more than one ways to do it. I think it would be good if these files are always generated. If we build this in it will also help users generate geometry headers for their own draggers. 

I know of a tool called ivToInclude that does the conversion on Windows as well, I have called it from a batch file when I did my own draggers years ago. It is hosted on Source Forge: http://sourceforge.net/p/inventor-tools/code/HEAD/tree/trunk/. There is a wiki as well: http://merlin.fit.vutbr.cz/wiki/index.php?title=Open_Inventor_Tools#ivToInclude. There may be a chicken-and-egg problem though as it uses Coin itself.

But conversion is actually not complicated. Looking at the script Coin/data/draggerDefaults/iv2h.sh.in it is basically a regexp replacement (stringification of the .iv file) wrapped in some boiler plate and an individualised include guard. Reading a bit about CMake [1] all of this could possibly be implemented using just CMake, without any other scripts or tools.

[1] http://www.cmake.org/pipermail/cmake/2007-August/015936.html

VolkerEnderlein:

We are not the first to do this. This is one ref, there are probably others:
https://github.com/Kitware/VisCL/blob/master/CMake/viscl-stringify.cmake

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

I'll investigate this and update the pull request. Thanks a lot!

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

You're welcome! Of course it's possible to pack the whole CMake commands into one single script file to avoid multiple copies. But I like the idea of splitting the stuff into multiple parts. This makes it easy for parallel development on multiple parts. That's also one of the reasons why I have deleted coin_doxygenfiles.cmake.inc. It's possible to define code and documentation for each part separately. This pull request is a first shot on that topic. I'll remove the duplications by a single list. For compilation I prefer only one project instead of 42 parts, but I created COIN_BUILD_SINGLE_LIB as a switch to turn it on or off - so feel free to stay with so many libs.

VolkerEnderlein:

Great! Let me know if you need help interpreting the sed command in the script. The s command is documented here [1], regular expressions here [2] and the 3 in there seems to be a line address [3].

[1] https://www.gnu.org/software/sed/manual/html_node/The-_0022s_0022-Command.html#The-_0022s_0022-Command

[2] https://www.gnu.org/software/sed/manual/html_node/Regular-Expressions.html#Regular-Expressions

[3] https://www.gnu.org/software/sed/manual/html_node/Addresses.html#Addresses

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Next steps:

* Generating code from *.iv
* Adding PDB options for Visual Studio debugging
* Remove the double file definitions in CMake scripts

VolkerEnderlein:

There shouldn't be a need to differentiate between shared and static if the same objects are to be linked. That is you can leave out the SHARED and STATIC words and have one add_library that serves both, if I read the [docs](http://www.cmake.org/cmake/help/v2.8.12/cmake.html#command:add_library) correctly: "If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is true."

VolkerEnderlein:

You might want to change to [LIST(APPEND ...)](http://www.cmake.org/Wiki/CMake_Performance_Tips#Use_LIST.28APPEND_....29) for better performance.

VolkerEnderlein:

What is the reason why you use an explicit list of sources for compilation and linking, but do a file(GLOB...) for documentation? Can we use the same list for both? We could glob for compilation as well, but an explicit list has the advantage that you can quickly copy a file to make a local backup during development without affecting the build.

VolkerEnderlein:

Would

```
#!CMake

set (COIN_VERSION "${COIN_MAJOR_VERSION}.${COIN_MINOR_VERSION}.${COIN_MICRO_VERSION}${COIN_BETA_VERSION}")
```
work?

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

file(GLOB ...) is collecting all files. If you clean up all directories of not used files, we can do this. The documentation is not complete yet. Some files should not parsed by doxygen.

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

No, because LIST is not working with global variables.

VolkerEnderlein:

Are you sure? "Similar to the SET command, the LIST command creates new variable values in the current scope, even if the list itself is actually defined in a parent scope. To propagate the results of these operations upwards, use SET with PARENT_SCOPE, SET with CACHE INTERNAL, or some other means of value propagation." http://www.cmake.org/cmake/help/v2.8.12/cmake.html#command:list

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

PARENT_SCOPE is only for parent, it won't work you have other non-parent script that want to see it as well. You need cache for the true "global-like" variable.

VolkerEnderlein:

Thanks. I am a bit out of my league here, I must admit.

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Don't worry about that ;-) I had the same idea and was wondering why it was not working.

VolkerEnderlein:

Sorry if I wasn't clear: I'd rather not use GLOB, so things work even when directories are not completely clean.

As for undocumented files: maybe we should discuss this, but I'd rather see all public interfaces appear in the documentation even when they are not documented. It is better than if they stay hidden. I don't mind Doxygen spitting warnings for undocumented members, but it can also be configured to not warn if we want this. I suppose we could use the same list as for compilation then?

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Ok, I'll change that.

VolkerEnderlein:

What is the reason for removing this? If left in, you don't need to test for COIN_BUILD_SHARED_LIBS on line 33 of src/CMakeLists.txt, add_library will do SHARED or STATIC based on the value of BUILD_SHARED_LIBS, right?

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

I make all project specific options with a COIN_ prefix. So if I include coin as a dependency in an other project I can differentiate between them. It's helpful if I like to build a shared library (DLL) with static coin (LIB).

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

yes

VolkerEnderlein:

I agree with that.

VolkerEnderlein:

Wow guys. You have been busy while I have been away! Just got back so I will have to spend a little time catching up.

Do you want me to merge this pull request yet?

Roy

VolkerEnderlein:

I don't understand the problem. Will BUILD_SHARED_LIBS leak to other projects if you depend on Coin? That would surprise me.

What I propose is

```
#!CMake

if (COIN_BUILD_SHARED_LIBS)
    SET(BUILD_SHARED_LIBS ON)
else()
    SET(BUILD_SHARED_LIBS OFF)
endif()
```

then you don't need the conditional when doing add_library sin src. 

From add_library docs: "If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is true."

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

I'll try to find a solution for the *.iv files. Do you like to wait for this?

VolkerEnderlein:

Bastiaan,

Having seen your explanation above I retract my earlier comment somewhat and agree with your idea. You are using a COIN_ prefix to control a built-in option. Presumably if, for example, we also had a simage project we could reproduce the same code with the SIMAGE_ prefix to set the option. It just means we always need to set the built-in option in case some usage has set it to the opposite of what was expected.

VolkerEnderlein:

I don't mind either way. Whatever works best for you guys.

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

I like to fix this first, also the RC file for Windows DLL is missing.

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

I've ignored the sed syntax and rebuild it by my own. But thanks your hints.

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Finished! Anything else to change?

VolkerEnderlein:

That's great,  transporter. I'll review it thoroughly tomorrow and if there are no objections I will merge it in.

Roy

VolkerEnderlein:

The second test for COIN_BUILD_SHARED_LIBS can be left out.
(Pardon to repeat myself, but my comment on the commit doesn't show on the actions page.)

VolkerEnderlein:

Excellent, well done!

VolkerEnderlein:
**Original comment by OgreT (Bitbucket: [OgreT](https://bitbucket.org/OgreT), GitHub: [OgreT](https://github.com/OgreT)).**

Yes, just testing for WIN32 is enough.

VolkerEnderlein:

This looks good to me. I wouldn't expect the scope of BUILD_SHARED_LIBS to become global after adding a project dependency. But I can't say for sure.

VolkerEnderlein:

Awesome! Just run a successful build on Windows, including generating the shader and inventor include files. :-)

PR #229, opened 2019-12-23
Change install destination so that coin.dll goes to the bin folder. Quarter does the same thing. So, then building them both with the same install prefix they install their dlls consistently to the bin folder.

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/46) by Nikolay Fedorov (Bitbucket: [Unril](https://bitbucket.org/Unril), GitHub: [Unril](https://github.com/Unril)).**

Change install destination so that coin.dll goes to the bin folder. Quarter does the same thing. So, then building them both with the same install prefix they install their dlls consistently to the bin folder.


VolkerEnderlein:

Nikolay,

I have noted that Mario has included your pull request code here in pull request #232. The one difference is that he uses the bin folder irrespective of the value of COIN_BUILD_SHARED_LIBS. Do you have any comment on which is better?

Roy

VolkerEnderlein:
**Original comment by Nikolay Fedorov (Bitbucket: [Unril](https://bitbucket.org/Unril), GitHub: [Unril](https://github.com/Unril)).**

Roy,
his way is better. I was mistaken a little in this place. While building static libs Coind.pdb file is not generated at all. Instead vc$(PlatformToolsetVersion).pdb is created due to /Zi compile flag. I think that it may be better to change /Zi flag to /Z7 in that case to embed debug info right to the Coind.lib itself.

https://msdn.microsoft.com/en-us/library/958x11bc.aspx

VolkerEnderlein:

Thanks Nikcolay. Are you happy for me to decline and close this pull request?

VolkerEnderlein:
**Original comment by Nikolay Fedorov (Bitbucket: [Unril](https://bitbucket.org/Unril), GitHub: [Unril](https://github.com/Unril)).**

Yes, it's OK.

PR #238, opened 2019-12-23
Edited math-undefs.h

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/55) by Nikolay Fedorov (Bitbucket: [Unril](https://bitbucket.org/Unril), GitHub: [Unril](https://github.com/Unril)).**

Edited math-undefs.h
Visual studio uses cosf and other float math functions internaly in its headers. So it causes compilation errors when compiling.


VolkerEnderlein:

these cases

VolkerEnderlein:

Because of a recent change to SbSphere.cpp (which now uses sqrtf), it is currently not possible to compile Coin with VS older than 2015.

Actually I think that most users should nowadays use a Compiler supporting C99, so maybe we can remove math-undefs.h completely. Otherwise we should fix SbSphere.cpp to not use sqrtf and fix the compiler warning using a static cast instead.

VolkerEnderlein:

I agree with Mario and now think this file should be removed. This mechanism to trigger errors if floating point versions are used internally no longer works with newer compilers and is therefore not effective in preventing the usage of those functions in Coin. Furthermore, floating point math functions are supported across platforms and compilers now for more than a decade. Mario or Nikolay, can create a separate pull request for removing this file?

VolkerEnderlein:
**Original comment by Markus Rickert (Bitbucket: [rickertm](https://bitbucket.org/rickertm), GitHub: [rickertm](https://github.com/rickertm)).**

Hi Mario,

I also agree with removing math-undefs.h as I think using undef on standard functions is not a good solution. It took me a while to find this issue when switching to VS 2015. It also seems the only intention behind this was to force the programmer to avoid using cosf, sinf?

As SbSphere.cpp is C++ code, you can also use std::sqrt instead of sqrtf, which is overloaded for float and double.

VolkerEnderlein:

Thank you for your feedback.

@walroy, @veelo: If you don't object, I'll merge this pull request to the default branch and remove the file from the CMake branch.

VolkerEnderlein:

No objections. Does that mean that MSVS 2015 or newer is a requirement on windows now? Or are we going to make other changes in support of older compilers?

VolkerEnderlein:

I think Markus is correct with his analysis. The intention was to force the developers to use the overloaded functions from C++ on a compiler level.

The real error is in SbSphere.cpp which should have used the C++ overloaded function sqrt instead of sqrtf in the commit.
If this is corrected older compilers will still work as intended. I checked it in the CMake branch. SbSphere.cpp is the only source where a C99 function from this header is used.

Edit: In case my previous comment was misleading, Yes, I also second that this header should be removed. But nevertheless should SbSphere.cpp be changed to use generic C++ function sqrt.

VolkerEnderlein:

When I read the comment in math-undefs.h, the original intention was to prevent the usage of float math functions, because
they were not standardized before C99.

The problem is, that math-undefs.h is included in almost every header file (indirectly, through SbBasic.h), so the float functions will
also be undef'ed in all client code, causing potential conflicts with other libraries.

Moreover the mechanism does not work anymore, because if someone now develops using VS >= 2015, he will not get compile errors, because of a change to math-undefs.h on the CMake branch.

I assume that all modern compilers support the float functions now. According to Nikolay's comment below, VS supports it since 2005. 
I tested with VS2008 and 2013 - sqrtf compiled without problems.
According to https://gcc.gnu.org/c99status.html, GCC support C99 "substantially completely" since V4.5 (released April 14, 2010).

Thus I think that math-undefs.h causes more trouble than it helps and should be removed now.

VolkerEnderlein:
**Original comment by Markus Rickert (Bitbucket: [rickertm](https://bitbucket.org/rickertm), GitHub: [rickertm](https://github.com/rickertm)).**

I've added corresponding pull requests to remove math-undefs.h from the default branch (pull request #269) and the CMake branch (pull request #270). Tested on Visual Studio 2010, 2017 and Ubuntu 16.04.

VolkerEnderlein:

Thank you Markus. @Renreok, will you do the honours?

VolkerEnderlein:

Markus, thank you very much for your work. I merged both pull requests and also removed the math-undefs.h file from both branches. I tested again with VS2008 on the defautl branch and 2017 on the CMake branch.

@veelo, @walroy: Do you know if it is possible to merge and close a branch in one commit using the bitbucket web UI as it is possible on the command line? I didn't know what would happen to the pull requests (whether they get marked as merged) if I merged the branches on the command line, so I decided to merge in the web UI and close the branches in two additional commits.

All: Thank you for the discussion and please let me know if you find any issues.

VolkerEnderlein:

Hi Mario,

It is possible, and mostly,  but not always, works. Sometimes it causes problems, so I always tend to do it in two stages, so as not to risk messing things up. The problem has been around for a couple of years, and has been reported. It has happened to me, and I see occasional references to it continuing to happen. So I would recommend merging first, then  closing the branch in a separate operation.

Roy

VolkerEnderlein:

Thank you all, Markus and Mario in particular.
@Renreok There is a checkbox saying something like "close branch" in the merge web UI, is that what you are referring to? There have been issues using that in the past, don't quite remember what the issue was. Maybe something like the branch being closed even if the merge failed or something, causing a bit of a panic. But I have used it on branches of my own with the intended result.

PR #251, opened 2019-12-23
proposal to add priority to setGrabber (handleEventAction)

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/68) by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

This branch implements a priority feature for the setGrabber method (similar to priorities in sensors). This allows to prioritize certain nodes when the grabber is set. Each node can apply a priority when requesting the grabber. The highest priority receives the grabber (with 100 the lowest priority to 0 the highest priority). By default the priority is set to 100. Since the priority has a default parameter, nothing changes for code bases that would not use this feature.

* added priority to setGrabber in handleEventAction

* added implementation for priority on setGrabber


VolkerEnderlein:
**Original comment by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

Roy, have you looked at this pull request? Any obstacles to merging it?

VolkerEnderlein:
**Original comment by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

If you would not like to merge these changes, can we change the releaseGrabber and setGrabber methods to virtual, so that they can be correctly overwritten in derived objects?

VolkerEnderlein:

Can you point to the priority functionality in sensors that you refer to? I cannot find it.

It would be interesting to see a use case for this, but I don't see a problem that would argue against merging; after you documented the feature --- that should come first.

I don't see a problem with making said functions virtual either.


Bastiaan.

VolkerEnderlein:
**Original comment by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

The priority functionality in the sensors I refer to is set by the function void SoDelayQueueSensor::setPriority (uint32_t  pri	). 0 is an immediate callback, higher number have lower priority.
The use case for using priorities for the grabber is when you want to change the priority of the grabber of nodes in the scene graph. Right now the last node wins, with priorities, the first node can win, if it uses a higher priority. So we can change which nodes receives the grabber without reordering them.
I would be happy to add some clarifying comments to the merge request.

VolkerEnderlein:

Thanks for the clarifications. I would accept this proposal -- if others have objections: let us know. Could you please change (or comment on) the following:

1. Please change the type from ```int``` to ```uint32_t```, like in ```SoDelayQueueSensor```.
1. I am not sure about using ```const``` for the priority argument. I know some people swear by it, but I don't think there is precedent for it in Coin.
1. By "documenting" I don't mean clarifying comments, but extending the Doxygen section above the function definition. It should explain what the argument does and how you can use it to achieve the objective you described above. And what the default and highest priorities are, and ideally also what happens if two nodes are of equal priority. All this will end up in the online API documentation.
1. The PR is conflicting at the moment, probably due to line endings. It should be Unix endings.

Thanks,
Bastiaan.

VolkerEnderlein:

I think it would be nice to have a more flexible way to determine which scene graph node should start interaction (e.g. in some cases it might be desired to rotate the camera, in other cases it might be desired to start dragging an object).

However, my feeling is that this pull request could cause problems on other ends, especially where SoDragger stores the pointer to the action where it set the grabber.
Did you check what happens, if SoDragger does not get the grabber (because some other node set the grabber with higher priority)?
From reading the source code, it seems that the SoDragger would still release the grabber in the destructor (even though it did not really get it). Can this lead to problems?

VolkerEnderlein:

This has been open for a while now and I propose to reject the pull request. I share the concerns that Mario raised. Even though Christoph made sure that the interface change is backwards compatible, I’m afraid that this has unforeseen side effects.

VolkerEnderlein:

Agreed, my requests have haven’t been addressed either.

VolkerEnderlein:
**Original comment by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

Abandoning my change request seems to be the consensus from our side as well.

VolkerEnderlein:

OK Christoph, declining then. Thank you anyway. If ever someone decides to have another look at implementing similar functionality then your code can be found in the “declined” bin for inspiration. Thanks also to the reviewers!

PR #254, opened 2019-12-23
CMake

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/71) by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

Creating pull request for a colleague, so there might be some delay in answering any question.

Some files are deleted since there are findings in them related to an open source copyright scan


VolkerEnderlein:

Well the review of this PR certainly is delayed -- sorry for that.

I advise against merging this PR for two reasons:

1) The request is a pull from the CMake branch of the fork to the default branch of upstream, essentially merging the CMake branch into default. Although that is something that should be done in the near future, it should not be done in this PR.

2) The actual content of this PR are these two commits: [6bf6ba0](https://bitbucket.org/chrvetter/coinjamesfix/commits/6bf6ba0e2f050dfbacb984604ada30c74ce64e4f) and [a255d90](https://bitbucket.org/chrvetter/coinjamesfix/commits/a255d90ddbc9744ca8437223736e4929844b690a), which supposedly solve issues regarding copyright.

It would be much more helpful if your colleague clarified what the issues actually are, rather than bluntly deleting files. It is my impression that these changes only benefit one person without knowledge of coding who thinks there is a legal danger in the presense of these files on his systems. Firstly I don't believe there is such danger (see the COPYING file) and secondly I don't believe these commits benefit anyone else.

Now some of these files may actually have lost their relevance and we could consider to delete some of them, but that should be a PR made out of a different assessment and with a different argumentation.


Anyway, thanks for your earlier contributions to Coin.

Bastiaan.

Issue #139, opened 2017-12-06
Notify in SoNode is not performed after SoNode nextUniqueId overflow.


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/139) by Peter Zarnay (Bitbucket: [PeterZarnay](https://bitbucket.org/PeterZarnay), ).**

----------------------------------------

It is allowed to overflow nextUniqueId which is int32, but after overflow notify is not called anymore and SoNode::getNodeId returns last value. There is condition in SoNode::notify where SoNotList timestamp has to be greater than uniqueId of particular node. But in case that nextUniqueId is overflowed it has minimum unsigned integer value again and this condition will not be met and particular node will not change its uniqueId anymore.

void
SoNode::notify(SoNotList * l)
{
#if COIN_DEBUG && 0 // debug
  SoDebugError::postInfo("SoNode::notify", "node %p (%s \"%s\"), list %p",
                         this, this->getTypeId().getName().getString(),
                         this->getName().getString(), l);
#endif // debug

  // only continue if node hasn't already been notified.
  // The time stamp is set in the SoNotList constructor.
  if (l->getTimeStamp() > this->uniqueId) {
    SET_UNIQUE_NODE_ID(this);
    inherited::notify(l);
  }
}

Dikluwe:
This was fixed shortly after being reported: commit b46eb13d0 ("Optionally use 64 bit integer type for unique node id (issue #139)", 2017-12-06) followed by 13dfc34f1 a few days later making 64-bit the default in all configurations. `SoNode::uniqueId`/`nextUniqueId` and `SoNotList`'s timestamp are now `SbUniqueId`, which defaults to `uint64_t` (see `include/Inventor/C/basic.h.cmake.in`) -- only reverting to the original 32-bit overflow behavior if an application explicitly opts back in via `COIN_UNIQUE_ID_UINT32` for ABI compatibility. At 64 bits, the overflow described here is no longer a practical concern. Closing as fixed.

PR #273, opened 2019-12-23
Optionally use 64 bit integer type for unique node id (issue #139)

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/90) by Thomas Moeller (Bitbucket: [TheHubbit](https://bitbucket.org/TheHubbit), GitHub: [TheHubbit](https://github.com/TheHubbit)).**

Peter and I looked into issue #139 and believe that the best solution is to use a larger identifier on 64 bit platforms. We added a new typedef SbUniqueId (similar to SbBool) that can be 32 or 64 bit depending on the CMake option COIN_UNIQUE_ID_64 (SbUniqueId remains uint32_t in all basic.h files not generated by CMake).


VolkerEnderlein:

Not that I have a problem with this proposal, but what speaks against making the ID 64 bit always, on all platforms?

VolkerEnderlein:

Coin supports so many and some very old configurations that we are just a bit cautious. Doing a quick search it seems though that pretty much all compilers support 64 bit integers even on 32 bit platforms. Therefore I agree with you that we should make COIN_UNIQUE_ID_64 available regardless of 32/64 bit compilation. The change to 64 bit ID potentially affects people who wrote their own nodes. Therefore we thought it is best to provide a configuration option to use 32 bit ID for backwards compatibility, at least for now. But I’m OK with removing that option and define SbUniqueId as uint64_t everywhere if we are all fine with that.

VolkerEnderlein:

Hi Thomas,

I like the idea of having a 64 bit node ID, but I think a lot more changes are necessary to really make it work. A quick search in the Coin solution for "getNodeId" returned 63 hits, and I think we need to have a look at all of them. E.g. elements use the node ID to quickly determine whether a cache is still valid (see e.g. SoAccumulatedElement::addNodeId). We should be able to find out if we were successful, if the number of warnings is the same in a build with 64 bit node IDs and a build with 32 bit node IDs.

I would also recommend to have a build option to switch between 32 bit and 64 bit node IDs, because similar adaptions might be necessary in client code.

Mario

VolkerEnderlein:

Thanks Mario for your comments! I'll update the pull request.

VolkerEnderlein:

I just updated the pull request addressing the comments from Bastiaan and Mario:

* SbUniqueId is a typedef to uint64_t in all build configurations unless COIN_UNIQUE_ID_UINT32 is defined

* SbUniqueId is now used consistently throughout Coin. No more warnings about data loss during conversion from SbUniqueId are shown when I compile the library.

VolkerEnderlein:

Thank you for updating the pull request. I reviewed it and it looks good to me now. I also compiled it and get the same number of warnings as before.

PR #274, opened 2019-12-23
SoBaseKit::getPart() sometimes creates parts even if makeIfNeeded is FALSE

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/91) by Mario Koerner (Bitbucket: [Renreok](https://bitbucket.org/Renreok), GitHub: [Renreok](https://github.com/Renreok)).**

If a SoBaseKit contains another SoBaseKit as part, then it can happen that parts are created when calling the getPart() method, even if makeIfNeeded is FALSE.

This can be reproduced with the following Code snippet:

```
#!c++
	SoShapeKit* sk = new SoShapeKit;
	SoNode* ak = sk->getPart("appearance", FALSE);
	assert(ak == nullptr); // ak must be NULL, because appearance is "null by default" part

	SoNode* mat = sk->getPart("material", FALSE);
	assert(mat == nullptr); // mat must be NULL, because appearance and appearance.material are "null by default"

	ak = sk->getPart("appearance", FALSE);
	assert(ak == nullptr); // expectation: ak is still NULL

```

When calling getPart for the "material" part (which is a part of the SoAppearanceKit "appearance"), then Coin internally creates an instance of the "appearance" part to search it for the "material" part. This is because Open Inventor node kits support partial part names, i.e. it is not necessary to specify the full part name as "appearance.material". However, if "makeIfNeeed" is FALSE, I don't see a need to search all non-existing node kit parts.

Another side effect of this bug is that when a node kit is searched for a part that does not exist at all in a specific node kit type, then getPart() will continuously create and destroy all node kit parts, causing e.g. sensors on the parent node kit to be fired. This is especially annoying when searching for a part in the sensor callback...

This pull request fixes the bug. I have it now running in my application for quite a while, but did not see any unexpected side effects so far. It would be nice if someone could review this fix, because it is a change at a rather central location.


VolkerEnderlein:

I would also expect the behavior after your change and therefore approve.

PR #253, opened 2019-12-23
turning off alpha test for small text quality improvement

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/70) by Christoph Vetter (Bitbucket: [chrvetter](https://bitbucket.org/chrvetter), ).**

----------------------------------------

alpha test removes parts of small characters resulting in bad text quality

Issue #126, opened 2016-10-12
Coin3D text quality issue from SoText2 node


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/126) by Xiaoyong Yang (Bitbucket: [yangxy0723](https://bitbucket.org/yangxy0723), GitHub: [yangxy0723](https://github.com/yangxy0723)).**

Attachments: [with alpha test.bmp](https://github.com/coin3d/attachments/blob/master/Coin3D_coin/issue126/with%20alpha%20test.bmp) | [without alpha test.bmp](https://github.com/coin3d/attachments/blob/master/Coin3D_coin/issue126/without%20alpha%20test.bmp)

----------------------------------------

In SoText2 node, alpha test is enabled with threshold of 0.3.  As a result, some fonts (e.g., time new roman) with small font size (16 in the attached example) have lower quality (comparing to FTGL) for rendering. The question is why we need alpha testing here?  If we disable the alpha test, the text rendering quality is as good as FTGL.

PR #275, opened 2019-12-23
Optimized rendering performance for SoText2

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/92) by Mario Koerner (Bitbucket: [Renreok](https://bitbucket.org/Renreok), GitHub: [Renreok](https://github.com/Renreok)).**

This pull request provides a performance optimized version of the SoText2 node.

The original implementation uses at least two GL calls per glyph: One call to glRasterPos3f() and one call to glDrawPixels() (for non-monochrome
fonts) or glBitmap (for monochrome fonts). These many GL calls are very inefficient with modern graphics cards and drivers, independent of whether
they are invoked directly or through a display list (using a SoSeparator with render caching enabled).
My optimization first constructs one bitmap with the whole text in main memory and displays it using a single set of GL calls, which turns out
to be much more efficient. So far my implementation only optimizes non-monochrome text, because this is the more important case for me.
Monochrome text still uses the old approach.

I used a SoText2 node with the first four strophes of Goethe's poem "Erlkoenig" with "Times New Roman" font in size 18 to
measure the performance on two different machines running Windows 10:

* [Laptop] Intel i5-3317U CPU (2.38 GHz) with Intel HD Graphics 4000 onboard
* [Workstation] Intel Xeon E5-1650 v3 (3.57 GHz) with Nvidia Quadro K5200

The results are as follows:

Version | [Laptop] | [Workstation]
---------|----------|---------
Original version without render caching|57 ms|3.1 ms
Original version with render caching|56 ms|2.6 ms
Optimized version without render caching|1.7 ms|0.4 ms
Optimized version with render caching|1.2 ms|0.2 ms

The rendering results with the optimized version are similar, but not binary equal with the original version. In my optimized version I always
round the text position to align it with the pixel grid, because interpolated text always looks ugly, and I don't see a reason not to round it.
If I modify the original version to do the same rounding and compare the result with my optimized version, then the result is binay equal in many
scenarios.

However, there are still a few differences in some scenarios, but I think they are acceptable:

* text with CENTER justification can be shifted by one pixel in horizontal direction, because my implementation rounds twice, once to draw
the glyph into the bitmap and once more to draw the bitmap on the screen
* the "greater than 30%" alpha function is only applied to the whole text in my implementation, while it is applied to every glyph in
the original version. This can cause small differences in pixel values where the glyph bitmaps of two glyphs overlap in a non-trivial manner
(i.e. both are non-zero at a certain position)
* order of alpha blending is different: while my implementation first blends all glyphs into one bitmap, the original implementation blended
every glyph directly over the background. Again this can make a few pixels appear slightly different where glyph bitmaps overlap in non-trivial
manner.

I found one more issue which is wrong in my implementation, but was also wrong in the original implementation. In some languages
(e.g. arabic) text should be printed right to left, but Coin always prints it left to right. Since this bug already existed, I did not fix
it with this pull request.

I have my version running in my application which dispays a lot of text for some time now. So far it seems to be stable, but it would
be nice if someone could try it in another application.


VolkerEnderlein:

Sorry I just pulled pull request #253, which I think conflicts with this one. While you resolve the conflict, could you keep an eye on issue #126 in order not to regress that?

Unless @chrvetter wants to review this, I think you can just merge this when you feel confident. You have done a thorough job evaluating and the performance increase is worth it.

VolkerEnderlein:

The images in issue #126 are not very meaningful, because it seems that the text was upscaled by a factor of about 3 (assuming that a font size of 16 was used for rendering), but with different Interpolation modes. @yangxy0723, @chrvetter maybe you can comment on how those Images were created.

However, I also do not understand .3 threshold for text rendering, and also created some Images with font "Times New Roman", size 16:



Even though I don't see a big difference, I would say that the right result is slightly better and will update my pull request accordingly.

VolkerEnderlein:
**Original comment by Xiaoyong Yang (Bitbucket: [yangxy0723](https://bitbucket.org/yangxy0723), GitHub: [yangxy0723](https://github.com/yangxy0723)).**

The original text was rendered by xipbuilder.  I did cut-paste the string rendering images using SnagIt.

VolkerEnderlein:

@Renreok Do you still want to update this PR? Otherwise I'd say merge it, so we can merge the CMake branch into default. This PR is the only one remaining for this branch. We can restore pull request #253 in default if needed.

VolkerEnderlein:

I will merge it now.

I already merged pull request #253 into this branch. However, pull request #253 is causing some problems in my application. If e.g. a shader is active that does not handle alpha blending,
then the text will now always be rendered with black background. Before the merge of pull request #253 the transparent pixels were simply rejected.
But since this is not directly related to the performance optimization, I'll merge my pull request now. We can fix that after the merge to the default branch.

PR #276, opened 2019-12-23
Make sure to continue sensor queue processing after unhandled exception

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/93) by Mario Koerner (Bitbucket: [Renreok](https://bitbucket.org/Renreok), GitHub: [Renreok](https://github.com/Renreok)).**

If a sensor callback function throws an exception which is caught by the code which called one of the sensor queue processing functions
of SoSensorManager, then the application stays alive, but will not process sensors of this type anymore. This is because
SoSensorManager sets an internal flag while processing a sensor queue which will not be reset in case of an exception.

This patch makes sure that the flag is reset also in case of an exception. It makes sure that SoSensorManager is left in a valid state when
a sensor callback throws an exception (e.g. all locks are released, all queues are in a valid state). The remaining sensors will then
be executed during the next call of the corresponding queue processing function.

Furthermore this patch logs an error message immediately when a sensor callback function throws an exception. This problem is then much
easier to investigate than follow-up problems which are caused by sensors that were not executed anymore.


VolkerEnderlein:

I think this is a good thing to have. Tanks!

PR #346, opened 2019-12-23
Let CMake handle prefix of include path

VolkerEnderlein:
**[Original pull request](https://bitbucket.org/Coin3D/coin/pullrequests/163) by Florian Franzen (Bitbucket: [Florian Franzen](https://bitbucket.org/Florian Franzen), ).**

There is no need to add the prefix here as relative path are always assumed relative to the prefix by CMake.

 Adding the prefix here however leads to an incorrect include path if absolute paths are being used.


VolkerEnderlein:

Thank you very much for the patch, it is highly appreciated.

PR #368, opened 2020-01-20
Fixed bug (crash) when call to cc_flww32_get_bitmap() fails

Renreok:
I would like to ask you to review a change in the Coin font handling.

Currently, Coin will crash when it fails to read a glyph either from the freetype library or from Win32. Although there is some error handling in place (Coin reads the glyph from the built-in default font), there are two problems:
- it does not mark the glyph as "fromdefaultfont", which causes several functions to fail (e.g. cc_flw_get_bitmap_advance() will return invalid information about how to advance to the next glyph in a text) and it will crash when it tries to free the glyph bitmap memory (fontstruct_rmglyph)
- glyphs from the default font use a different reference line than other fonts. The default font uses the descender line, while other fonts usually use the base line.

The first problem can be solved by properly setting the "fromdefaultfont" flag. I did quite some code review and testing, and I think that setting this flag truely solves the problem.

The second problem can be solved by setting the "bearing" for the default font in the same way as for fonts that use the freetype library or Win32 API. Note that this change will change the rendering output of any Coin program that uses the default font. However, I think it makes sense to have consistent behavior of all fonts.

Here is some Illustration, how text will be rendered before and after my fix (default font is rendered on the left, while "Times New Roman" is rendered on the Right).

old
new

P.S. The reason why I need this fix is that in my project we observe sporadic issues with the GetGlyphOutline Win32 function, which fails sporadically. I'm still trying to find the root cause for the failure of this function, but I think it still makes sense to fix the issues with the Coin font handling.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,9 @@
-New in Coin v4.0.0 (so far):
+New in Coin v4.1.0 (so far):
+* bugfixes:
+  - fixed crash when reading font glyphs fails
+  - use font base line as bearing point for default font
+
+New in Coin v4.0.0 (2019-12-23):
 * new:
   - Kongsberg Oil & Gas Technologies AS ended Coin as a commercial product
     and re-licensed it under the BSD 3-clause license as a service to the
--- a/RELNOTES
+++ b/RELNOTES
@@ -8,6 +8,16 @@ The changes described here are issues that might have an effect on the
 compatibility of your source code when upgrading the library.
 
 
+Coin 4.1.0 (so far)
+==========
+
+Reference line for text rendered with the built-in default font was changed
+(before: descender line, now: base line). Thus text rendered with default
+font will appear a few pixels lower than before, but now consistent
+with text using non-default fonts.
+(see https://github.com/coin3d/coin/pull/368)
+
+
 Coin 4.0.0
 ==========
 
--- a/src/fonts/builtin2dfonts.icc
+++ b/src/fonts/builtin2dfonts.icc
@@ -11,6 +11,11 @@
 #define COIN_FONT_25_HEIGHT (27)
 #define COIN_FONT_33_HEIGHT (36)
 
+#define COIN_FONT_13_BEARING (12)
+#define COIN_FONT_17_BEARING (15)
+#define COIN_FONT_25_BEARING (21)
+#define COIN_FONT_33_BEARING (28)
+
 static uint8_t font_data_13[256][COIN_FONT_13_HEIGHT*4] = {
 #include "font13.icc"
 };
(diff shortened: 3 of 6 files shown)


veelo:
In principle, I support the change to base line positioning. It is consistent, and I think it is what people expect.

Hypothetically, there could be an application that lines out labels with objects (like the annotation of dimensions in a technical drawing), and in this case the current positioning helps legibility because the slight vertical separation is independent of zoom level (but dependent on font size). I am not aware of a way to offset text dependent on font size, so I think this application would have trouble attaining the same effect.

It would not affect any of my projects negatively (and I do use text). Would this be a problem for FreeCAD, @kkremitzki?

veelo:
Anyway, before this gets accepted, please add a clear notice in https://github.com/coin3d/coin/blob/master/RELNOTES so that users are properly informed of the change.

Thanks!

kkremitzki:
It would be good to do some testing of this with FreeCAD, but I may not be able to get to it before the weekend. I will create a thread in the FreeCAD forums asking if anyone is available to test this to make sure nothing breaks.

Renreok:
Thank you very much for your replies. It is not urgent to merge this change, it should be a deliberate decision.
It would also be an option to introduce a compiler switch, but too many build switches can also be confusing. To me the inconsistent behavior feels more like a bug, so my initial proposal is without build switch.

Thanks for the hint about the RELNOTES file, Bastiaan. I really like that we maintain release notes now.

Renreok:
Im currently working on the update of the RELNOTES file. Before I commit, I'd like to ask about the Coin version we are now working on (4.0.1?) and whether I should also adapt the NEWS file.

kkremitzki:
IMO, no need for the NEWS file for this sort of change, and the current version should be 4.1.0.

veelo:
> Im currently working on the update of the RELNOTES file. Before I commit, I'd like to ask about the Coin version we are now working on (4.0.1?) and whether I should also adapt the NEWS file.

I'd like you to add a line to the NEWS file, with a reference to RELNOTES for details and possibly this PR (for the images).

We haven't started working with different branches yet for different versions, which we probably should consider when changes pick up pace. According to https://semver.org/ I'd say the change in positioning is a minor change, i.e., targeting 4.1.0. If only the crash were addressed then 4.0.1 would be more appropriate, but IMO there is no need to be pedantic about this.

Renreok:
I updated the pull request with changes to the NEWS and RELNOTES files. Please have a look.

Regarding branching: I think we should decide about the version we are targeting with the master branch (it seems currently 4.1.0), and also update the version defines (e.g. in config.h) accordingly.

Renreok:
Please let me know if you need more time to analyze the Impact of this change.
If there are no objections, I will merge this pull request at the end of this week.

PR #385, opened 2020-05-10
Reverting commit 8ef77a076472feb995b7c9690a534e2e6e749a75 to restore correct rendering for SoText2.

### LINKED ISSUE #384: Render issue in SoText2 because of missing alpha test
We draw small coordinate systems with axis labels and noticed a rendering regression in Coin 4.0.0 vs. 3.1.3 (see the attached images). The transparency modes we are using are `SORTED_OBJECTS_BLEND` and `SORTED_LAYERS_BLEND`.
AlphaTestErrorCoin4 0 0
AlphaTestOKCoin3 1 3

After close inspection it looks like commit 8ef77a076472feb995b7c9690a534e2e6e749a75 introduced this faulty behaviour.

VolkerEnderlein:
Reverting commit 8ef77a076472feb995b7c9690a534e2e6e749a75 to restore correct rendering for SoText2. fixes #384

### DIFF
--- a/src/shapenodes/SoText2.cpp
+++ b/src/shapenodes/SoText2.cpp
@@ -419,6 +419,7 @@ SoText2::GLRender(SoGLRenderAction * action)
     glPushAttrib(GL_ENABLE_BIT | GL_PIXEL_MODE_BIT | GL_COLOR_BUFFER_BIT);
     glPushClientAttrib(GL_CLIENT_PIXEL_STORE_BIT);
     
+    SbBool didenableblend = FALSE;
     SbBool drawPixelBuffer = FALSE;
 
     for (int i = 0; i < nrlines; i++) {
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for your comments Mario. My knowledge about glyph handling in Coin is very limited, so my first attempt was simply to revert the commit in question. But I agree that the code can and should be simplified. So do I understand you correctly, that **all** glyphs of a SoText2 will be either drawn with a glBitmap call (per glyph) or by a call to glDrawPixelBuffer (for the combined glyphs in the pixel_buffer)?

Renreok:
Yes, that's how it works. While the original implementation drew the text glyph by glyph (and potentially toggled between glBitmap and glDrawPixelBuffer), I refacored it (starting with commit 609653f, pull request #275) to first copy all non-monochrome glyphs into a pixel buffer and render it in one step at the end. However, I did not change the rendering of monochrome glyphs, because fonts are usually grey-scale today and I did not want to implement all the necessary bit operations to copy the monochome bitmap into the pixel buffer.

That's why the original implementation needed "didenableblend", but the latest version does not.

PR #404, opened 2020-07-21
Fix GLX glue trying to use NULL display

### LINKED ISSUE #401: Replace GLX with EGL for Wayland
This is a suggestion from the FreeCAD forum.
FreeCAD crashes because of GLX enabled. 
The workaround includes disabling GLX in coin-hg.
https://forum.freecadweb.org/viewtopic.php?t=33359#p351418

myfreeweb:
i.e. when a unix platform has a non-GLX context: usually EGL-Wayland, but also might be offscreen or maybe EGL-X11.

This patch fixes #401 (well, it doesn't introduce EGL glue, but people just want FreeCAD running on Wayland).

### DIFF
--- a/src/glue/gl_glx.cpp
+++ b/src/glue/gl_glx.cpp
@@ -236,10 +236,9 @@ static Display *
 glxglue_get_display(const cc_glglue * currentcontext = NULL)
 {
   if (currentcontext && currentcontext->glx.glXGetCurrentDisplay) {
-    if (glxglue_screen == -1) {
-      glxglue_screen = XScreenNumberOfScreen(
-	XDefaultScreenOfDisplay(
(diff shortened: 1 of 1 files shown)


myfreeweb:
> Err:6 http://azure.archive.ubuntu.com/ubuntu bionic-updates/main amd64 libegl-mesa0 amd64 19.2.8-0ubuntu0~18.04.3 404  Not Found [IP: 52.154.174.208 80]

hmm looks like your CI needs to `apt-get update` before upgrade/install

VolkerEnderlein:
Thanks for notifying us about the issue and for providing the fix. It is very much appreciated.

kkremitzki:
It might be nice to have a release which includes this bugfix, say 4.0.1 if appropriate.

PR #416, opened 2020-10-15
Fix various typos/grammar

luzpaz:
Found via `codespell`


veelo:
Thanks a lot! These error are eye hurting...

PR #452, opened 2021-02-27
Fix for issues found with PVS Studio static analysis

VolkerEnderlein:


veelo:
Thank you for your tireless and meticulous work on Coin3D, Volker!

PR #465, opened 2021-08-23
Basic helloworld exmaple

magres:
Could be useful for newbies.

### DIFF
--- a/examples/helloworld/README.md
+++ b/examples/helloworld/README.md
@@ -6,6 +6,6 @@ Building instructions:
 
     cd <source path of example>
     mkdir build && cd build
-    cmake -DQt5_DIR=<path of Qt5> -DCoin_DIR=<path of Coin3d lib> -DSoQt_DIR=<path of SoQt lib> ..
+    cmake -DQt5_DIR=<Qt5 path> -DCoin_DIR=<Coin3d path> -DSoQt_DIR=<SoQt path> ..
     cmake --build .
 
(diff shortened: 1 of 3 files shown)


examples/helloworld/README.md
```diff
@@ -0,0 +1,9 @@
+# HelloWorld example
+
+Very basic helloworld example to get started with Coin3d. 
+
+Building instructions:
+
+    cmake -DQt5_DIR=<Qt5 path> -DCoin_DIR=<Coin3d path> -DSoQt_DIR=<SoQt path> <example source path>
```
VolkerEnderlein, line 7: According to this CMake [question](https://stackoverflow.com/questions/56581343/why-doesnt-find-package-look-in-name-dir-lib-cmake-name-version) on StackOverflow you should use either `<package name>_ROOT` variables as command line arguments or define the `<package name>_DIR` variables as environment variables. That's why I mostly recommend using the `CMAKE_PREFIX_PATH` argument on the command line call to CMake (for instance: `cmake -S <sourcedir> -B <builddir> -DCMAKE_PREFIX_PATH="<Qt5 path>;<Coin3d path>;<SoQt path>" -G "<selected CMake generator>`) You might omit the `<SoQt path>`, as in general SoQt will be installed in the Coin3d path.

The CMake build command should explicitly mention the config:  `cmake  --build <builddir> --config Release`

VolkerEnderlein:
Thanks for the example. I'll happily merge your PR if you adapt the build instructions.

luzpaz:
bumping

examples/helloworld/README.md
luzpaz: @magres :point_up_2:

veelo:
Superseded by #632.

PR #469, opened 2021-10-03
Rework the ERROR definition problem with Win32

### LINKED ISSUE #468: <Inventor/errors/SoDebugErrors.h> leaves macro ERROR defined to an undeclared symbol
The header defines `ERROR` as an undeclared symbol. It seems it tries to save and restore the original definition of `ERROR` but it fails to do so.

https://github.com/coin3d/coin/blob/dfb56d294eeef3581e4322ed8881f478fc242dec/include/Inventor/errors/SoDebugError.h#L85

kazssym:
This change will solve the name collision problem with the Win32 API.

I hope it is safe to redefine `ERROR` as an enum constant instead of a preprocessor macro.

Fixes #468.

### DIFF
--- a/include/Inventor/errors/SoDebugError.h
+++ b/include/Inventor/errors/SoDebugError.h
@@ -39,8 +39,12 @@
 // Avoid problem with Microsoft Win32 API headers (yes, they actually
 // #define ERROR -- in wingdi.h).
 #if defined(ERROR)
-#define SODEBUGERROR_STORE_ERROR_DEF ERROR
+enum {
+  SODEBUGERROR_ERROR_VALUE = ERROR,
 #undef ERROR
+  ERROR = SODEBUGERROR_ERROR_VALUE,
+};
+#define ERROR ERROR
 #endif /* ERROR */
 
 
@@ -79,11 +83,5 @@ private:
   static void * callbackData;
   Severity severity;
 };
-
-// Avoid problem with Microsoft Win32 API headers (see above).
-#if defined(SODEBUGERROR_STORE_ERROR_DEF)
-#define ERROR SODEBUGERROR_STORE_ERROR_DEF
-#undef SODEBUGERROR_STORE_ERROR_DEF
-#endif /* SODEBUGERROR_STORE_ERROR_DEF */
 
 #endif // !COIN_SODEBUGERROR_H


include/Inventor/errors/SoDebugError.h
```diff
@@ -39,8 +39,12 @@
 // Avoid problem with Microsoft Win32 API headers (yes, they actually
 // #define ERROR -- in wingdi.h).
 #if defined(ERROR)
-#define SODEBUGERROR_STORE_ERROR_DEF ERROR
```
VolkerEnderlein, line 42: Rather than defining an additional enum I would use the following sequence in the `SoDebugError` header and source files:

```
#if defined(ERROR) && defined(_WIN32)
#pragma push_macro("ERROR")
#undef ERROR
#endif

< code that defines enum in header and that uses enum in source >

#if defined(_WIN32)
#pragma pop_macro("ERROR")
#endif
```

This will save the value of the `ERROR` macro on the stack with the `push_macro` command, undef the macro in the current context and restore its previous value after the `pop_macro` command.
Although `push_macro` and `pop_macro` are Microsoft preprocessor extensions (defined at least since VS6 from 1998) they are also widely supported from in Intel C++, GCC and Clang compilers.

VolkerEnderlein (commented):
Thanks for your contribution to Coin development. Please see the comment attached to your PR.

include/Inventor/errors/SoDebugError.h
kazssym: It looks good, but `SoDebugError::ERROR` will become an error then?

VolkerEnderlein: Not in `SoDebugError.h` and `SoDebugError.cpp`, the `ERROR` preprocessor token is undefined and a new enum constant `SoDebugError::ERROR` can be defined without conflicts. But in other sources that eval `SoDebugError::ERROR` **and** include `windows.h` you need to insert the same sequence. A quick grep in the entire Coin3d repository revealed only one place in `SoWin.cpp` where `SoDebugError::ERROR` is used. This `push_macro`/`pop_macro` is used for instance in Google protobuf (see [here](https://sourcegraph.com/github.com/protocolbuffers/protobuf/-/blob/src/google/protobuf/port_def.inc) )

Issue #505, opened 2023-08-14
SoOffscreenRenderer problems when running on RHEL8 and SLES12


shanslow:
I'm not sure if this is a problem per se, or if there's advice that can be given on something we may be doing incorrectly, but we're currently seeing issues outputting image files to disk on RHEL8 and SLES12. Our software is fully built on RHEL7.9 but deployed on these sister versions. Historically that's worked fine building on version X and running on X+1 from RHEL4 onwards but seemingly no longer for RHEL7->RHEL8.

Rendering to the screen is fine, but when we instantiate and the SoOffscreenRenderer and attempt to render to file we get the following messages before the appliaction segementation faults:

_Coin warning in glxglue_context_create_software(): Couldn't create GLX context.
Coin warning in SoOffscreenRenderer::renderFromBase(): Could not set up an offscreen OpenGL context._

The problem seems to manifest at line 959, with the call to _this->glcanvas.deactivateGLContext();_ where it attempts to deactivate the offscreen GL canvas that it's created. This ultimately ends up in _glxglue_context_reinstate_previous_ where the _glXMakeCurrent_ call fails. On RHEL7 it appears to be able to store a reusable context in the _glxglue_contextdata_ but the same information is null on RHEL8.

Is this a problem or could someone point out what we may be missing to get the offscreen rendering feature to work for us? Our build flags for building Coin, and the source below, are minimal with nothing specific other than optimisation level, ensure it's 64 bit and position independent code. We initially thought it may be a VM related issue with a software video driver, but the problem is manifesting on real hardware with dedicated GPU.

The attached source below exhibits the problem, it renders (badly!) a simple cuboid scene, which it then attempts to write to disk, which works without issue On RHEL7.

Thanks

[source.txt](https://github.com/coin3d/coin/files/12332609/source.txt)

ytterx:
Hi,

We are encountering the same issue, it seems it is due to that on RHEL8 "Indirect GLX" is disabled.
Advisory on Xorg:
https://www.x.org/wiki/Development/Security/Advisory-2014-12-09/
It might be possible to turn it on in your case in `/etc/X11/xorg.conf`.

I'm very curious if a code change to coin can fix this.

shanslow:
[...]

PR #507, opened 2023-10-01
Adding environment variable COIN_GLX_PIXMAP_DIRECT_RENDERING to control direct rendering option for offscreen contexts

ytterx:
Proposed fix for #505
Probably needs more testing on various systems and configurations

### DIFF
--- a/src/doc/Coin_environment.dox
+++ b/src/doc/Coin_environment.dox
@@ -268,6 +268,7 @@
   \li \ref COIN_CGLGLUE_NO_PBUFFERS
   \li \ref COIN_GLXGLUE_NO_PBUFFERS
   \li \ref COIN_GLXGLUE_NO_GLX13_PBUFFERS
+  \li \ref COIN_GLX_PIXMAP_DIRECT_RENDERING
   \li \ref COIN_WGLGLUE_NO_PBUFFERS
 
   \li \ref COIN_ALLOW_SPIDERMONKEY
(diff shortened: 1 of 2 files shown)


ytterx:
@VolkerEnderlein Hi, you seem to be the most recent contributor that merged a pull-request. Is is possible for you to take a look at this issue?

VolkerEnderlein:
Probably something linke that:

```
Bool direct_rendering = False;
static int check_direct = -1;
if (check_direct == -1) {
  check_direct = 0;
  const char * env = coin_getenv("COIN_GLX_PIXMAP_DIRECT_RENDERING");
  direct_rendering = env && atoi(env) == 1 ? True : False;
}

context->glxcontext = glXCreateContext(display, context->visinfo, 0, direct_rendering);
```

ytterx:
That might be a workable approach. I was hoping I could detect that it would not work and then allow for an automated switch but I wasn't able to do something like that. I'll make the suggested changes to my Pull-Request and test it in our environment.

VolkerEnderlein:
I just noticed, the direct_rendering variable must be static too. Otherwise it works only for the first time but not for subsequent calls.

ytterx:
Made the required changes, also added the information to the documentation

VolkerEnderlein:
Thanks @ytterx for providing the necessary changes. Your contribution to Coin development is highly appreciated.

PR #509, opened 2023-11-06
Improve auto clipping

Rexbas:
While trying to fix an orthographic camera auto clipping problem in FreeCAD, I noticed that during camera motions, for both orthographic and perspective camera, the corners of a part are sometimes cut. See the attached video which is running at 5% the original speed.

https://github.com/coin3d/coin/assets/14298143/f2594dd8-e7a9-4223-9957-d32b4201d308

I think what happens is this: first the camera orientation and/or position is changed, then the scene is rendered, then the auto clipping runs. After the camera moved, the scene may no longer be completely in the viewing frustum. So there is maybe one frame that shows the cut corner before the auto clipping runs and renders the scene again. But if the camera is in motion then there are multiple consecutive frames and the cut corners are noticeable.

To fix this, I created 2 commits. The first commit runs the auto clipping before rendering the scene and this fixes the cut corner problem. Then I thought: if the auto clipping runs before rendering then it is no longer needed to use a clip sensor. So I removed the clip sensor and related methods in the second commit. I am not sure if I can just do that without breaking something for someone. Please let me know what you think :smiley:

### DIFF
--- a/include/Inventor/SoRenderManager.h
+++ b/include/Inventor/SoRenderManager.h
@@ -220,9 +220,7 @@ protected:
 
 private:
   void attachRootSensor(SoNode * const sceneroot);
-  void attachClipSensor(SoNode * const sceneroot);
   void detachRootSensor(void);
-  void detachClipSensor(void);
   static void nodesensorCB(void * data, SoSensor *);
   static void prerendercb(void * userdata, SoGLRenderAction * action);
 
--- a/src/rendering/SoRenderManager.cpp
+++ b/src/rendering/SoRenderManager.cpp
@@ -270,11 +270,6 @@ SoRenderManager::SoRenderManager(void)
 
   PRIVATE(this)->glaction = new SoGLRenderAction(SbViewportRegion(400, 400));
   PRIVATE(this)->audiorenderaction = new SoAudioRenderAction;
-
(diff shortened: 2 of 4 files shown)


VolkerEnderlein:
Thank you for your contribution to Coin development, it is highly appreciated. I'll check the PR for side effects on supported platforms and report back. Keep up the good work.

Rexbas:
@VolkerEnderlein have you had any time to review this? No hurry though, just wondering.

VolkerEnderlein:
@Rexbas unfortunately not. I was busy getting the coin 4.0.1 and soqt 1.6.1 done. I had a quick look but no time to test thoroughly.

Rexbas:
@VolkerEnderlein, I would really appreciate a review, do you think that is possible in the near future?

VolkerEnderlein:
I am sorry for the very long delay. I need to test applications that add additional clip planes to make sure this PR does not have side effects.

Rexbas:
@VolkerEnderlein, is there any way I can help with testing to make sure there are no side effects?

VolkerEnderlein:
Sorry for the very long delay. The PR looks good to me and has no side effects with my current test programs.
Thank you for your contribution to coin development. Keep up the good work.

PR #510, opened 2023-11-08
Fit entire scene between near and far plane for auto clipping orthographic camera

Rexbas:
In FreeCAD there is an issue (https://github.com/FreeCAD/FreeCAD/issues/6836) where faces of a part disappear or the part completely disappears when using the orthographic camera. The attached video shows the problem.

https://github.com/coin3d/coin/assets/14298143/8263c8d0-518f-4a38-bce5-24a5d4183b54

This problem has been partially fixed in FreeCAD 0.21. Changing the behavior of auto clipping for the orthographic camera in coin could completely fix this problem. It looks like this problem occurs in situations where the farDistance is positive and the nearDistance is negative.

In this PR, I fixed it by moving the camera back or forward such that the nearDistance is 1 and the farDistance is the depth of the bounding box of the scene + 1. The focal distance is changed to still have the same focal point.

I am not sure if the behavior as in the video above is the intended behavior or not. If it is the intended behavior then I could also change this PR to add a new AutoClippingStrategy. Please let me know.

### DIFF
--- a/src/rendering/SoRenderManagerP.cpp
+++ b/src/rendering/SoRenderManagerP.cpp
@@ -35,6 +35,7 @@
 
 #include <Inventor/nodes/SoInfo.h>
 #include <Inventor/nodes/SoCamera.h>
+#include <Inventor/nodes/SoOrthographicCamera.h>
 #include <Inventor/nodes/SoPerspectiveCamera.h>
 #include <Inventor/actions/SoGLRenderAction.h>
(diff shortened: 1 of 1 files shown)


ytterx:
Hi, it looks like the issue I was describing in https://github.com/coin3d/soqt/issues/70
I'll try to find some time next week to test this.

VolkerEnderlein:
Thanks a lot for your contribution to Coin development, it is highly appreciated. I'll check the PR for side effects on supported platforms and report back.

PR #518, opened 2024-01-29
Set minimum C++ standard needed to compile Coin

ric-bianchi:
The compilation of the Coin 'master' fails out of the box because the minimum C++ standard is not specified in the CMake configuration, and the old standard is picked by CMake by default (at least on macOS).

This PR adds the minimum C++ standard to CMakeLists.txt. 

That fixes the out-of-the-box build of Coin. 

Tested on macOS 13.5.1 + Xcode 15.1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,12 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
 endif()
 
 # ############################################################################
+# Set minimum C++ standard needed to compile the Coin code
+# ############################################################################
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the PR. I just checked with the macos-13 github runner image (macos-latest still defaults to macos-12). It looks like the dependency to boost 1.84 causes coin not to compile with C++ 03 anymore, as boost 1.82 declared support for C++03 deprecated.

VolkerEnderlein:
I am still unsure about merging the PR, as it is only a 3rd party dependency that requires this. So I started a poll on the Coin3D discussions tab https://github.com/orgs/coin3d/discussions/520 . Feel free to vote. Cheers.

luzpaz:
Any progress ?

PR #519, opened 2024-02-10
Fix orthographic autoclipping by allowing negative farDistance

Rexbas:
Replaces #510.  The `SoRenderManagerP::setClippingPlanes` would return if `farval <= 0.0f`. Now it does not return anymore if an orthographic camera is used. This fixes the problem described in #510 in a more simple way.

### DIFF
--- a/src/rendering/SoRenderManagerP.cpp
+++ b/src/rendering/SoRenderManagerP.cpp
@@ -36,6 +36,7 @@
 #include <Inventor/nodes/SoInfo.h>
 #include <Inventor/nodes/SoCamera.h>
 #include <Inventor/nodes/SoPerspectiveCamera.h>
+#include "Inventor/nodes/SoOrthographicCamera.h"
 #include <Inventor/actions/SoGLRenderAction.h>
 #include <Inventor/actions/SoGetBoundingBoxAction.h>
 #include <Inventor/actions/SoGetMatrixAction.h>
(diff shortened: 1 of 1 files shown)


src/rendering/SoRenderManagerP.cpp
```diff
@@ -133,7 +134,12 @@ SoRenderManagerP::setClippingPlanes(void)
   float nearval = -box.getMax()[2];
   float farval = -box.getMin()[2];
 
-  if (farval <= 0.0f) return;
+  if (!camera->isOfType(SoOrthographicCamera::getClassTypeId()) && farval <= 0.0f) return;
+
+  if (box.isEmpty()) {
+    nearval = 0;
```
VolkerEnderlein, line 140: If the box is empty I would set the default value for the nearval which is 1.0

VolkerEnderlein: Thanks for the PR, I run a simple test and it did not show any unwanted effects. After resolving this conversation I am happy to merge the PR.

Rexbas: Thanks, I updated the nearval to 1

VolkerEnderlein:
Thanks for your contribution to Coin development. It is highly appreciated.

luzpaz:
@VolkerEnderlein FreeCAD is coming up on a 1.0 Release milestone (https://blog.freecad.org/2024/08/30/1-0rc1-is-coming-next-week-lets-update-addons/). Is there any chance there can be a coin release beforehand so some of these fixes are available ?

PR #525, opened 2024-05-03
Modern documentation

githubuser0xFFFF:
This pull request adds modern documentation build target based on doxygen-awesome style


VolkerEnderlein:
Thank you Uwe for the hard work you put into modernizing Coin documentation. It is highly appreciated.

githubuser0xFFFF:
Thank you for merging this. How do you distribute the documentation? I looked at the Github actions but could not find a documentation task.

VolkerEnderlein:
There is no task as building the documentation has not been automated until now.

VolkerEnderlein:
I just added the necessary steps to the CI to automatically build and publish the documentation for Coin and Quarter. The links on coin3d.github.io / coin3d.org point now to this awesome documentation.

githubuser0xFFFF:
Thank you for Volker for adding this to CI - great. There is only a minor issue that you I see on mobile devices: If I create the documentation here with doxygen 1.10.0, the hamburger menu is properly working:

grafik

In the documentation that has been compiled by CI with doxygen 1.9.1, the hamburger menu does not work and the menu eats to much screen from screen:

grafik

So if it is possible in CI to use doxygen 1.10.0, then I would recommend to use this. If you do not have the time to do this, then I will look into this.

VolkerEnderlein:
Thanks Uwe, I'll have a look onto the issue tonight.

VolkerEnderlein:
Hi Uwe, I split the workflow into a documentation and a CI part. Then I updated the doxygen version to 1.10.0 but it needed to be compiled from source. Up to now this is done on every push. Do you know if we can cache that artifact and avoid the continuous rebuild of doxygen? Can you check whether the docs work properly on mobile devices? Cheers, Volker

githubuser0xFFFF:
Hi Volker. Documentation looks good now on mobile devices - great. I have no great experience with github workflows and don't know how to cache artifacts - sorry.

Another option is to use Ubuntu 20.04 for documentation build and just use the tarball instead of builder doxygen from source.

PR #533, opened 2024-10-30
Update repology badge(s)

AMDmi3:
coin4 was merged info coin3d in Repology, so two distinct badges are not longer needed


VolkerEnderlein:
Thank you for your contribution to Coin development, it is highly appreciated. Keep up the good work.

PR #541, opened 2024-12-25
Fix macro handling to avoid conflicts with Win32 API headers

### LINKED ISSUE #468: <Inventor/errors/SoDebugErrors.h> leaves macro ERROR defined to an undeclared symbol

kazssym:
Adjust macro definitions to prevent conflicts with Microsoft Win32 API headers by using push and pop macro directives.

Fixes #468

### DIFF
--- a/include/Inventor/errors/SoDebugError.h
+++ b/include/Inventor/errors/SoDebugError.h
@@ -39,7 +39,8 @@
 // Avoid problem with Microsoft Win32 API headers (yes, they actually
 // #define ERROR -- in wingdi.h).
 #if defined(ERROR)
-#define SODEBUGERROR_STORE_ERROR_DEF ERROR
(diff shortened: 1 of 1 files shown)


kazssym:
This PR changes the `SODEBUGERROR_STORE_ERROR_DEF` macro's usage as the marker to pop the definition of `ERROR`. Due to the language spec, it could not be used to restore the original definition at all.

VolkerEnderlein:
Thanks for reworking the prvious PR. Your contribution to Coin development is highly appreciated.

PR #549, opened 2025-02-10
EGL fixes and GLFW example

tritao:
This PR has some EGL fixes, along with changes to support both EGL and GLX simultaneously.

There is now a `COIN_EGL` env var that can be used to control which context to use.

It also introduces a `COIN_BUILD_EXAMPLES` build option, and a new GLFW-based example app.


VolkerEnderlein:
@tritao Thanks for the PR it is highly appreciated. For the broken merge checks I needed to change the CI workflow because one GitHub action was removed. But from the CodeQL workflow I could deduce the correct working of your PR. Thanks again.

Rexbas:
> along with changes to support both EGL and GLX simultaneously.

@tritao, how exactly does this work. If I use `-DCOIN_BUILD_EGL:BOOL=ON -DCOIN_BUILD_GLX:BOOL=ON` then `HAVE_GLX` gets commented out in config.h: `/* #undef HAVE_GLX */`.

It does work when I manually type `#define HAVE_GLX` but when I reload CMake it gets removed again.

Rexbas:
Yes, I was testing with both EGL and GLX for FreeCAD with the idea that if it works, a new version of coin could be released. That way at least all users who don't use NVIDIA proprietary drivers have an option to go use Wayland. Although not by default because then we may get angry NVIDIA users.

PR #550, opened 2025-02-13
More robust EGL display initilization.

tritao:
@VolkerEnderlein Before merging lets wait until folks confirm it fixes the issue.

Based on a patch by @lepalom.

### DIFF
--- a/examples/glfw/glfw.cpp
+++ b/examples/glfw/glfw.cpp
@@ -61,7 +61,6 @@ void exposeCallback(void)
   glEnable(GL_LIGHTING);
 
   sceneManager->render();
-
   glfwSwapBuffers(window);
 }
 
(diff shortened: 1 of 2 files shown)


Rexbas:
I still get this:
`Coin warning in cc_glglue_instance(): Error when setting up the GL context. This can happen if there is no current context, or if the context has been set up incorrectly.`

But now I also get this message. I did not get this previously with the NVIDIA proprietary drivers:
`libEGL warning: egl: failed to create dri2 screen`

I remember that without the NVIDIA proprietary drivers, I also got this message. So I think the PR might be some positive progress.

Also now I get a black FreeCAD instead of white: 
image

Aside from the dri2 screen message, Hello Cone is still the same (no cone).

lepalom:
Anton Gladky solved, at leat for Debian, could you check it?
https://salsa.debian.org/science-team/coin3/-/blob/master/debian/patches/Egl_dust_off2.patch?ref_type=heads
@tritao

heiserhorn:
@lepalom I am running Debian sid with the latest version of coin3 with the patch mentioned above but I have exactly the same problem. I have commented on the debian closed bug which should be reopened.

luzpaz:
soft bump

PR #557, opened 2025-04-01
update CMAKE min version

mosfet80:
cmake 3.0 is deprecated. 
You need to upgrade to 3.5 (minimum supported version in cmake 4.0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.10)
 
 set(COIN_MAJOR_VERSION 4)
 set(COIN_MINOR_VERSION 0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.0)
 
 set(COIN_MAJOR_VERSION 4)
 set(COIN_MINOR_VERSION 0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.5)
 
 set(COIN_MAJOR_VERSION 4)
 set(COIN_MINOR_VERSION 0)
--- a/src/coin-config.cmake.in
+++ b/src/coin-config.cmake.in
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.10)
 
 # Have a look at VTKConfig.cmake.in, especially regarding UseVTK.cmake
 # https://github.com/Kitware/VTK/blob/master/CMake/VTKConfig.cmake.in
--- a/src/coin-config.cmake.in
+++ b/src/coin-config.cmake.in
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.10)
 
 # Have a look at VTKConfig.cmake.in, especially regarding UseVTK.cmake
 # https://github.com/Kitware/VTK/blob/master/CMake/VTKConfig.cmake.in
--- a/src/coin-config.cmake.in
+++ b/src/coin-config.cmake.in
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.5)
 
 # Have a look at VTKConfig.cmake.in, especially regarding UseVTK.cmake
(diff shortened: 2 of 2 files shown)


VolkerEnderlein:
Thank you for the PR. I was not aware about the upcoming deprecation in CMake version 4.0. According to [CMake documentation](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features) a command `cmake_minimum_required(VERSION 3.0...3.29)` is the preferred way to work with both obsoleted and current versions. 3.29 is the last version I checked explicitly. If you'll change your PR to this line I happily accept it.

mosfet80:
> Thank you for the PR. I was not aware about the upcoming deprecation in CMake version 4.0. According to [CMake documentation](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features) a command `cmake_minimum_required(VERSION 3.0...3.29)` is the preferred way to work with both obsoleted and current versions. 3.29 is the last version I checked explicitly. If you'll change your PR to this line I happily accept it.

https://cmake.org/cmake/help/latest/release/3.31.html

[Deprecated and Removed Features](https://cmake.org/cmake/help/latest/release/3.31.html#id14)
Compatibility with versions of CMake older than 3.10 is now deprecated and will be removed from a future version.

VolkerEnderlein:
The CMake release notes [link](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features) and the `cmake_minimum_required` [documentation](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#cmake-minimum-required) exactly propose the syntax I mentioned for supporting both newer *and* older CMake versions.

mosfet80:
The link you cite states that cmake versions < 3.5 have been removed.
Boost version 1.83 also requires cmake 3.5.
https://github.com/boostorg/boost/blob/boost-1.83.0/CMakeLists.txt

In order to support old and new version of cmake I propose to update to version 3.5

Thank you for your availability to review

mosfet80:
I would like to point out that when github compiling "C.I." on ubuntu, it fails because of the incompatibility with cmake 4.

CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

https://github.com/coin3d/coin/actions/runs/14204262087/job/39798140761

One of these decisions can be made to solve the problem:
# not to support cmake 4 
# impose cmake3.5 as the minimum version.
# add  into cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5

VolkerEnderlein:
The failing calls to CMake tell three ways how to solve the issue. 

```
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

And instead of raising the minimum required version I prefer the **`<min>...<max>`** syntax to keep backwards compatibility.

oursland:
I have attempted to alter the line to `cmake_minimum_required(VERSION 3.0..3.29)` as suggested and it results in the following output:

```
CMake Error at .pixi/envs/default/lib/cmake/Coin-4.0.3/coin-config.cmake:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
Call Stack (most recent call first):
  cMake/FreeCAD_Helpers/SetupCoin3D.cmake:12 (find_package)
  CMakeLists.txt:109 (SetupCoin3D)
```

It would appear that the minimum version does need to be at 3.5 or above, which was confirmed locally.  @mosfet80's PR appears necessary.

VolkerEnderlein:
You need to add three `.` between min and max numbers so it must read `3.0...3.29` to work properly. The `cmake_minimum_required` documentation mentions the exact reasons for this somewhat unusual syntax.

oursland:
@VolkerEnderlein Oh geez, yes the missing `.` was responsible for the failure.  My use of `cmake_minimum_required(VERSION 3.0...3.31)` functions as expected.

VolkerEnderlein:
@tritao @oursland @mosfet80  I am going to apply the proposed changes in a different PR as there was no timely reaction of the OP and the FreeCAD dev is stalled.

VolkerEnderlein:
The latest CMake video had a section how to prepare your package to support CMake 4.0. It advised to use exactly the syntax I was preferring.
[See here](https://www.youtube.com/watch?v=TESpxsilA1k&ab_channel=Kitware) at 27:35

cxwx:
It's time to mv to cmake4.0

PR #568, opened 2025-06-05
Introduce shaded hidden lines visual style

tetektoza:
Since we have different visual styles available in coin, but we lack visual style that allows showing hidden edges with solid edges, this patch adds it.

Made on FreeCAD request, this could be done directly in ViewProviders, although every viewprovider is being assigned separately to an object one after another, and since objects are being rendered one after depth buffer of obj1 may not be aware of obj2 being infront of the object, if obj2 was created after obj1. So, the solution was to add it to RenderManager as the scene will be conscious about depth of all objects.

Looks like this on the demo:

https://github.com/user-attachments/assets/342e4663-65d0-4a35-ab5b-937917445420

### DIFF
--- a/include/Inventor/SoRenderManager.h
+++ b/include/Inventor/SoRenderManager.h
@@ -84,7 +84,8 @@ public:
     POINTS,
     WIREFRAME_OVERLAY,
     HIDDEN_LINE,
-    BOUNDING_BOX
+    BOUNDING_BOX,
+    SHADED_HIDDEN_LINES
   };
 
   enum StereoMode {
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Sure, I'll do that in the upcoming week.

VolkerEnderlein:
Thank you for your contribution to Coin3D development it is highly appreciated.

VolkerEnderlein:
All has built but 32bit failed for some obscure reasons on MSYS2. The issue is on the CI side not in your code. Keep up the good work.

PR #571, opened 2025-06-12
Add sdl3 example

captain0xff:
This PR adds an example to demonstrate how to use coin3d along with the SDL3 library. The code is based on the glfw example.
I have only tested this on linux.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory(dynloadextensions)
 add_subdirectory(glfw)
+add_subdirectory(sdl3)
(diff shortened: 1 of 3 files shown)


VolkerEnderlein:
Yes, definitely, but I did not found the time to test and review your contribution properly.

VolkerEnderlein:
I'll be happy to merge your contribution to Coin3D if you resolve the ony issue I mentioned in my comment. Thanks for your valuable contribution. It is highly appreciated.

captain0xff:
I have addressed the issue. Apart from that, I have renamed the example from sdl3 to sdl3-example as a executable named sdl3 in the bin folder felt a bit wrong.
Thank you for the review :)
Coin3d is a very interesting library and I will definitely keep sending more patches.

Issue #556, opened 2025-03-27
Crash in SoCallbackAction


dhoepfl:
When looking at [this issue](https://github.com/FreeCAD/FreeCAD/issues/20303), I noticed something in SoCallbackAction which might be a memory leak (but then realized it is not but nonetheless explains the crash) - I only read the source, I did not compile or debug it. So please bare with me if this report is incorrect:

Calling [`SoCallbackAction::addPreCallback`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L520) calls [`set_callback_data`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L504) which in turn calls [`set_callback_data_idx`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L491).

This extends the given [`SbList`](https://github.com/coin3d/coin/blob/master/include/Inventor/lists/SbList.h#L70) to have enough entries, adding NULL entries as necessary, then places a pointer to a `new`ly allocated instance of [`SOCallbackData`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L226) in the appropriate entry.

Subsequent calls to `addPreCallback`, use [`append`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L232) of the first `SOCallbackData` to add allocated instances of `SOCallbackData`s.

After a few calls, the list looks like this:

```
list[0]->NULL
…
list[idx-1]->NULL
list[idx]->[SOCallbackData 1]-next->[SOCallbackData 2]-next->[SOCallbackData 3]-next->NULL.
```

The [destructor of `SoCallbackAction`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L471) calls [`delete_list_elements`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L462) on the list, which calls for each index [`deleteAll`](https://github.com/coin3d/coin/blob/master/src/actions/SoCallbackAction.cpp#L238) on the element it points to (`SOCallbackData  1` in the above example). It also tries to call `deleteAll` on entries that were set to NULL when expanding, thus calling `((SoCallbackAction *)NULL)->deleteAll()` which leads to crashes.

VolkerEnderlein:
Thanks for the thorough bug report and your analysis. I'll try to figure out what's needed to prevent the crash.

[...]

PR #572, opened 2025-06-19
Issue #556 Only call deleteAll on valid pointers

shanslow:
Adding safety check to only call deleteAll on callback actions that have been registered.

### DIFF
--- a/src/actions/SoCallbackAction.cpp
+++ b/src/actions/SoCallbackAction.cpp
@@ -462,7 +462,11 @@ static void
 delete_list_elements(SbList<SoCallbackData *> & cl)
 {
   int n = cl.getLength();
-  for (int i = 0; i < n; i++) cl[i]->deleteAll();
+  for (int i = 0; i < n; i++) {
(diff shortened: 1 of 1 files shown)


shanslow:
Is the Windows build something I'm able to address? It looks like it was unable to find boost on the build machine.

VolkerEnderlein:
Thanks for the fix. Your contribution is highly appreciated.

> Is the Windows build something I'm able to address? It looks like it was unable to find boost on the build machine.

No, I am aware of the issue and need to change the installed Boost from a source dist that was sufficient for older CMake versions to a binary dist that provides BoostConfig.cmake files. This changed when GitHub was updating their CMake installation in the windows-latest image. https://github.com/coin3d/coin/commit/40ee7eb38463da37cce30763a5b127597f92faa5 was a partial try but did not fix the root cause. I need to figure out what exactly to do including downloading, installing and caching the binary Boost dist for latest VS compiler and haven't done until now.

PR #576, opened 2025-07-02
Only use GLX on platforms with X11.

oursland:
Currently GLX libraries are being added to targets on platforms without X11, resulting in build failures.  This PR limits GLX to platforms running X11.


VolkerEnderlein:
Thanks for the fix and your contribution to Coin development.

PR #577, opened 2025-07-05
Add a clipping offset

Rexbas:
When the near and far camera distance are small and close to each other then things may not get rendered properly. For example https://github.com/FreeCAD/FreeCAD/issues/22216, here a sketch is created which is just some stuff on a 2D plane and the near and far distance are very close to each other because the depth is very small. Coin adds some slack, I believe 0.1% which is a multiplication and thus the amount of slack depends on the values of the near and far distance. However, if the near and far distance are close to 0 then effectively no slack is added. This could cause hidden objects and flickering when moving the mouse or camera.

To resolve this I propose we subtract an offset from the near value and add an offset to the far value. This way there is always sufficient distance between the near and far clipping plane.

### DIFF
--- a/src/rendering/SoRenderManagerP.cpp
+++ b/src/rendering/SoRenderManagerP.cpp
@@ -131,8 +131,9 @@ SoRenderManagerP::setClippingPlanes(void)
   xbox.transform(mat);
   SbBox3f box = xbox.project();
 
-  float nearval = -box.getMax()[2];
-  float farval = -box.getMin()[2];
+  constexpr float clippingOffset = 0.1;
+  float nearval = -box.getMax()[2] - clippingOffset;
+  float farval = -box.getMin()[2] + clippingOffset;
 
   if (!camera->isOfType(SoOrthographicCamera::getClassTypeId()) && farval <= 0.0f) return;
 


VolkerEnderlein:
I have some concerns about using an absolute value for the clippingOffset. Would it not be more appropriate to set the clippingOffset based on the depth of the bounding box (i.e. 0.1*depth)? That way we don't lose precision when dealing with very small objects. And we can set a lower limit on the offset, in case the bounding box becomes too small. The case of an empty bounding box is handled later so this should suffice.

Rexbas:
The problem is that sometimes there is no depth, or a very small value of depth. For example a 2D plane viewed along the normal of the plane. This may cause floating point accuracy issues in the calculation of the near and far clipping plane and thus it may not be rendered properly (flickering or partially hidden). Using a fixed offset that does not depend on the bounding box should make sure there is always sufficient distance between the near and far clipping plane, to both sides of the object, to properly render thin objects. Besides, we already have slack which is a multiplication but this is ineffective with small values.

VolkerEnderlein:
If there's no depth, at least one of the other extensions width and or height should have a value that can be used to adjust the clipping offset relatively to its size. Sure, it fails if the plane is kilometers long, millimeters wide and micrometers thick, but in the general case we would not limit the clipping volume to a minimum of 0.2 (nearval = -0.1, and farval = +0.1), what do you think?

Rexbas:
That may work, I will do some experimenting this week and let you know.

Rexbas:
I am now using 1% the length of the diagonal of the bounding box, and clamping it between 1 and float epsilon. This is working great.

VolkerEnderlein:
Thanks for the fix. It is highly appreciated.

VolkerEnderlein:
I added a small fix. instead of using powf(2, x) which caclculates 2^x using powf(x, 2) would be required. But using powf in this case is not neccessary and can easily be replaced by x*x which is much more performant in debug builds. The optimizer will do the right thing and optimize the function call away, but it's better to formulate it efficiently from the beginning.

Rexbas:
Nice!

PR #599, opened 2026-01-27
Remove legacy OpenGL fallback definitions

tritao:
## Summary
Remove the bundled legacy OpenGL fallback enums/macros and adjust a few call sites accordingly.

## Why
`include/Inventor/system/gl.h` had accumulated a large set of OpenGL/extension fallback definitions for very old headers. Keeping those in-tree increases maintenance and can hide portability problems by making missing symbols “appear” at compile time.

## Changes
- Slim `include/Inventor/system/gl.h` down to a small wrapper that defers to `Inventor/system/gl-headers.h`.
- Update a handful of OpenGL-related sources to use the proper enum/extension names without relying on the removed fallback definitions.

#### Commits
- `07e361d208`: GL: add enum fallbacks for legacy headers

### DIFF
--- a/include/Inventor/system/gl-fallbacks.h
+++ b/include/Inventor/system/gl-fallbacks.h
@@ -1,0 +1,689 @@
+#ifndef COIN_GL_FALLBACKS_H
+#define COIN_GL_FALLBACKS_H
+
+/* Fallback enum values for legacy OpenGL/GLU/WGL headers.
+ *
+ * Some environments (notably Windows) still ship very old OpenGL headers
+ * where common enums are missing unless additional extension headers are
+ * available and included.
+ *
+ * This list is intentionally limited to symbols known to be missing in
+ * such environments. Values are taken from the historical fallback block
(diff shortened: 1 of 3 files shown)


tritao:
This PR ended up a bit different than the original version, but before I complete it so it's ready for review, maybe it's worth having an initial discussion.

In a branch I've been working on, I end up adding support for the GLAD extension loader, which turns out some of this work obsolete, as it can fully replace the incomplete platform-specific headers and legacy fallbacks in the existing code or in this branch.

Do you think replacing the existing legacy OpenGL definitions with GLAD (https://glad.dav1d.de/) is an acceptable path going forward? @VolkerEnderlein

WizzerWorks:
There are also these links for reference:

- https://[gen.glad.sh](https://gen.glad.sh/)/
- https://github.com/Dav1dde/glad

VolkerEnderlein:
I just had a closer look onto GLAD, it looks well suited for the job. That path looks very promising.

PR #617, opened 2026-03-28
Fixes and improvements

captain0xff:
This PR is part of the work to vendor coin with FreeCAD and there are three commits that do the following:
1. Introduces a alias target Coin::Coin when building coin in tree as a sub-directory. This allows us to link the same target irrespective of building in tree or through find_package.
2. Fixes compilation with C++ 20 as std::shared_ptr::unique was [removed](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique.html) in C++20 in favour of use_count.
3. Updated the code to explicitly use ASCII versions of winapi functions. So, made GetCurrentDirectory -> GetCurrentDirectoryA. Functions like GetCurrentDirectory are just a macro that either expand to GetCurrentDirectoryA or GetCurrentDirectoryW (the unicode variant). We can undefine the _UNICODE macro in freecad to force the winapi macros to expand to the ASCII variant but I think keeping it explicit is better as the coin code doesn't compile with the unicode variants. I primarily develop on linux so not 100% confident on this change and can revert this commit if requested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,14 +555,18 @@ if(WIN32)
     set(COIN_DEFAULT_POSTFIX ${COIN_DEFAULT_STATIC_POSTFIX})
     set(COIN_LIBRARY_SUFFIX ${CMAKE_STATIC_LIBRARY_SUFFIX})
   endif()
-  set(CMAKE_RELEASE_POSTFIX "${COIN_DEFAULT_POSTFIX}"
-    CACHE STRING "Default filename postfix for libraries under configuration Release")
-  set(CMAKE_MINSIZEREL_POSTFIX "${COIN_DEFAULT_POSTFIX}"
-    CACHE STRING "Default filename postfix for libraries under configuration MinSizeRel")
-  set(CMAKE_RELWITHDEBINFO_POSTFIX "${COIN_DEFAULT_POSTFIX}"
-    CACHE STRING "Default filename postfix for libraries under configuration RelWithDebInfo")
-  set(CMAKE_DEBUG_POSTFIX "${COIN_DEFAULT_POSTFIX}d"
-    CACHE STRING "Default filename postfix for libraries under configuration Debug")
+  # Don't set these if the project is built as cmake a subdirectory
+  if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    set(CMAKE_RELEASE_POSTFIX "${COIN_DEFAULT_POSTFIX}"
+      CACHE STRING "Default filename postfix for libraries under configuration Release")
+    set(CMAKE_MINSIZEREL_POSTFIX "${COIN_DEFAULT_POSTFIX}"
+      CACHE STRING "Default filename postfix for libraries under configuration MinSizeRel")
+    set(CMAKE_RELWITHDEBINFO_POSTFIX "${COIN_DEFAULT_POSTFIX}"
+      CACHE STRING "Default filename postfix for libraries under configuration RelWithDebInfo")
+    set(CMAKE_DEBUG_POSTFIX "${COIN_DEFAULT_POSTFIX}d"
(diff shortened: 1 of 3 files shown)


captain0xff:
Ah, the PR from @chennes was merged just now. I was wondering how I got a conflict because I rebased quite recently.

VolkerEnderlein:
Thanks for the PR. I would not recommend to use the...A versions of the functions as this affects the entire set of users. It simply  prevents them to build a Unicode aware version. I am OK with the other changes.

captain0xff:
> I would not recommend to use the...A versions of the functions as this affects the entire set of users. It simply prevents them to build a Unicode aware version. I am OK with the other changes.

The problem is that the UNICODE aware version doesn't compile with the current codebase. We pass `const char*` but the UNICODE aware versions expect `wchar_t*` so the code fails to compile.

For example see the docs for https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcurrentdirectory
It used LPTSTR which expands too:
image
And as I wrote above LPTSTR is a typedef for whar_t* https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/50e9ef83-d6fd-4e22-a34a-2c6b4e3c24f3

The reason why I bothered with this change is that FreeCAD enables UNICODE support but coin fails to build with it. I can disable unicode support in FreeCAD before adding coin as a subdirectory but the code will still fail for anyone that wants UNICODE. Personally I think it is better to be explicit than failing if someone wants UNICODE until the code is refactored. Unless I misunderstood your message.

captain0xff:
I created a PR enabling unicode support in my own repo to confirm this in a isolated environment with coin only (as previously I tested only with freecad) and as you can see the windows CI fails:
https://github.com/captain0xff/coin/pull/1

R-Goc:
That particular one is just passing a string literal to GetModuleHandle. In this case just patch it to always use the wide version and add L to the literal. It has no impact on users at all. There might be other things uncovered later in the build or testing though.

captain0xff:
You can see my 3rd commit to see all the required changes that stop the build. Also, adding L will cause failures if someone wants to build without UNICODE support, right?
There are multiple errors in the above CI too for example:
image

R-Goc:
No, L just means it's a wchar_t literal. You can have those without UNICODE defined. Unicode affects the TEXT() macro and the expansion of functions without a suffix that have a unicode and ansi variant. In general using the W version when possible is the direction the ecosystem has gone in, and dealing with TCHAR is error prone. Also the system itself has to convert to WCHAR anyway as that's used internally.

captain0xff:
No, I mean the wchat_t version will not compile if the macro expands to the A version of the functions because they expect a `const char*`. So, I can't just make it behave properly with the W version.
The code will need separate branches for ansi and unicode and that is what I meant by proper refactor in the description.

captain0xff:
@VolkerEnderlein so what is the verdict on this one?

R-Goc:
> No, I mean the wchat_t version will not compile if the macro expands to the A version of the functions because they expect a `const char*`. So, I can't just make it behave properly with the W version.
> The code will need separate branches for ansi and unicode and that is what I meant by proper refactor in the description.

If you only use the W version, not the macro you can safely use wchar_t literals. If this is internal to coin and doesnt depend on runtime strings it is preferred to only use the wide versions. Otherwise you could also use the TEXT() macro but that is a bit ugly. Defaulting to ASCII is comparatively worse.

captain0xff:
In that case I am not the best person to do this as I don't even have a windows machine atm. I am dropping that commit and will try to patch the build issue from FreeCAD side.

VolkerEnderlein:
I can make and test the necessary changes as my primary dev environment is windows. Maybe you drop point 3 from this PR and I prepare a separate PR over the next days.

captain0xff:
Done!

VolkerEnderlein:
Thanks @captain0xff for your restless work on the Coin codebase. It is highly appreciated. Keep up the good work.

PR #623, opened 2026-06-13
Add pick layer to SoPickStyle for ordered on-top picking

### LINKED ISSUE #546: Rendering and picking nodes on top with correct depth support
To render nodes on top one could use `SoAnnotation`. This node however disables use of the depth buffer so nodes are rendered using rendering order, not depth order which is not always desirable. To circumvent that problem it is possible to simply split rendering into two (or more) steps separated by clearing of the depth buffer either by splitting the graph into two separate roots, or accumulating nodes into proper path lists and then dispatching the rendering action to them separately.

This works well for rendering, however `SoPickAction`s are not aware of such "steps" or "layers" and return incorrect order of picked points. Using `SoPickStyle::SHAPE_ON_TOP` results in objects being on top, which is problematic when there are multiple objects overlapping "on top" as they considered in rendering order. 

My proposal is to add additional `priority` property into `SoPickStyle` node and `SoPickStyleElement`. Elements would then be sorted first by their priorities (higher priority means that objects should be picked first), and then by the distance from camera. For compatibility, objects with `SHAPE_ON_TOP` would be considered as having highest possible priority with distance set to 0 (like it is now). This should would add only one more comparison to the sorting, which should not be problem in terms of performance.

Once agreed that it is ok, I can implement such feature. If wanted layered rendering could be implemented directly into Coin3D however because depth buffer would have to be cleared between each layer it could result in potentially changed behavior for code that relies on having proper depth buffer after dispatching the action rendering action.

kadet1090:
This will let picked points be ordered across the scene graph independent of depth. SoPickStyle gains a `layer` field (default 0), carried through traversal by the new SoPickLayerElement. SoRayPickAction sorts picked points by layer (descending) and then by distance (ascending), so a higher layer always picks in front while true depth order is preserved within a layer.

The existing *_ON_TOP styles are unchanged: they ignore the layer and still sort frontmost. This replaces the previous behaviour where on-top picks collapsed their distance to 0 as the only way to bias picking to the front.

Fixes: #546

### DIFF
--- a/include/Inventor/elements/SoElements.h
+++ b/include/Inventor/elements/SoElements.h
@@ -64,6 +64,7 @@
 #include <Inventor/elements/SoMaterialBindingElement.h>
 #include <Inventor/elements/SoNormalBindingElement.h>
 #include <Inventor/elements/SoPickStyleElement.h>
+#include <Inventor/elements/SoPickLayerElement.h>
 #include <Inventor/elements/SoSwitchElement.h>
 #include <Inventor/elements/SoTextOutlineEnabledElement.h>
 #include <Inventor/elements/SoTextureCoordinateBindingElement.h>
--- a/include/Inventor/elements/SoPickLayerElement.h
+++ b/include/Inventor/elements/SoPickLayerElement.h
@@ -1,25 +1,25 @@
-#ifndef COIN_SOPICKSTYLE_H
-#define COIN_SOPICKSTYLE_H
+#ifndef COIN_SOPICKLAYERELEMENT_H
+#define COIN_SOPICKLAYERELEMENT_H
 
 /**************************************************************************\
  * Copyright (c) Kongsberg Oil & Gas Technologies AS
  * All rights reserved.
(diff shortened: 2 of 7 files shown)


VolkerEnderlein:
Thank your for your valuable extension of Coin. It is highly appreciated. Can you rebase the PR on the HEAD, as I already merged two other PRs (including one that fixed the building issue on Windows OS) . Once the PR is ready to merge we need to bump the main version number as a new member variable is added to the existing class SoPickStyle and this breaks ABI interface.

veelo:
Sorry for interfering. I saw the rebase button, and forgetting that I wanted a minor release first, I had clicked the merge button...

veelo:
@kadet1090 let me know if you want me to take care of rebasing and opening a new PR with your commit; it was me who messed up.

kadet1090:
@veelo I was a bit busy and without access to computer to push it - but I've created #633 already. Thanks for the offer!

PR #624, opened 2026-06-14
Fixed some vulnerabilities reported by @oddmunds

VolkerEnderlein:


src/base/namemap.cpp
veelo: Should probably be `strncpy`.

VolkerEnderlein: Ahh, thanks, I missed that :)

src/base/namemap.cpp
```diff
@@ -153,7 +153,8 @@ find_string_address(const char * s)
     headchunk = newchunk;
   }
 
-  (void)strcpy(headchunk->curbyte, s);
+  (void)strncpy(headchunk->curbyte, s, headchunk->bytesleft - 1);
```
veelo: This pads the entire remainder of the chunk with 0. I don't know how much of an overhead that is in practice, and whether that is anything to worry about. I would probably have used `min(len, head hunk->bytesleft - 1)` but honestly C++ is not my main language. Just wanted to mention this.

veelo: Sorry len + 1

VolkerEnderlein: As far as i see the ending '\0' character is already included in len counting, so min(len, head_thunk->bytesleft) should suffice. Setting the last character to \0 after copying ensures that the character buffer is \0 terminated after copying, also in case the string is longer than the CHUNK_SIZE.

veelo: Agreed.

veelo (approved):

PR #625, opened 2026-06-16
Fix some vulnerabilities found by @oddmunds, GHSA-hcqw-f3m9-mq7p.

veelo:
Replaced fixed length buffers in parsing functions with a reused string of growing capacity.

Unfortunately we cannot remove the offending code completely yet, as it is used in the public API. I have instead made private copies and fixed those for use in the implementation, keeping as much of the original code as possible. Now all functions share a `std::string` to read into, which is `clear()`ed by each function and reused, keeping allocations low, if any. The unsafe API has been  deprecated, we should remember to remove that in a mayor release.

I have tested that this fixes the respective vulnerabilities, but I have not tested that file import still works as it should. @VolkerEnderlein I am kind of hoping that you have something readily available to verify this on. If not, I'll have to put more work in. Please let me know, I'll keep this in draft until confirmed.

As always, thanks for all the work that you put in!

Bastiaan.

### DIFF
--- a/src/coindefs.h
+++ b/src/coindefs.h
@@ -86,6 +86,9 @@
   information. Used where there is an obsoleted or unsupported
   function. Typically a function that we feel should have been private
   in Open Inventor.
+
+  COIN_DEPRECATED: Macro to be used as attribute of deprecated
+  function declarations. Emits warnings during compilation.
 */
 
 #if COIN_DEBUG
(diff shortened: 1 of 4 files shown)


veelo:
This also solves the additional finding Read Access Violation in Input Processing (SoInput.cpp:1252).

veelo:
I verified the correctness of floating point parsing.

PR #630, opened 2026-06-20
Deprecate SoInput::readHex(uint32_t & l).

veelo:
Forwards to SoInput::readUnsignedInteger(uint32_t & l), which supports a superset.

### DIFF
--- a/src/io/SoInput.cpp
+++ b/src/io/SoInput.cpp
@@ -867,27 +867,13 @@ SoInput::getASCIIFile(char & c)
   A number in hexadecimal format must have the "0x" prefix.
 
   Returns \c FALSE if end of file is encountered.
+
+  \deprecated Use readUnsignedInteger instead.
 */
 SbBool
 SoInput::readHex(uint32_t & l)
 {
-  assert(!this->isBinary());
-
-  // FIXME: this is a tremendously stupid function. Should obsolete it.
-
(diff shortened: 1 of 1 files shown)


veelo:
@VolkerEnderlein this should go into the security release for Coin 4, if possible. It fixes another potential buffer overrun.

VolkerEnderlein:
Sorry, I totally missed that one, when preparing the 4.0.10 release. Maybe we should create a 5.0 branch that will need some more work before publishing a new major release. Then Kacper Donats PR could be applied first and after that all the deprecations and interface changes. Before updating we would merge back the 5.0 branch to the master. What do you think about this?

veelo:
Yes I think that is a good idea, and we should have split off that branch on my inadvertent merge commit of Kacper’s PR, then he wouldn’t have had to redo it :-) We can still do that, then he doesn’t have to change the target of his new PR. On 21 Jun 2026, at 12:19, Volker Enderlein ***@***.***> wrote:﻿VolkerEnderlein left a comment (coin3d/coin#630)
Sorry, I totally missed that one, when preparing the 4.0.10 release. Maybe we should create a 5.0 branch that will need some more work before publishing a new major release. Then Kacper Donats PR could be applied first and after that all the deprecations and interface changes. Before updating we would merge back the 5.0 branch to the master. What do you think about this?

PR #632, opened 2026-06-20
helloworld example

veelo:
Revival of #465.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,3 +1,4 @@
 add_subdirectory(dynloadextensions)
 add_subdirectory(glfw)
+add_subdirectory(helloworld)
 add_subdirectory(sdl3)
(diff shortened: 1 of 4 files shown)


tritao:
Should we name this `qt` instead to match the other `glfw` and `sdl3` examples?

VolkerEnderlein:
Yes, I think that would be more appropriate.

veelo:
I have thought about that as well. The other examples have much more code though, so if someone is looking for a "Hello, World!" example, this fits the bill (not just the code, but also the cmake commands). Actually I am not that fond of the fact that it uses SoQt, as Quarter is supposed to replace SoQt. I have considered to replace it with a Quarter example, but that would be a replication of one of the existing examples in the Quarter repository. And this examiner viewer with fancy dials does not yet exist using Quarter (it could be ported I think, but although it looks interesting, it is not that useful in practice other than quickly testing simple scene graphs).

I would propose to keep this as is, and copy one appropriate example from the Quarter repository and name it Qt. This way we guide people familiar with Qt to the preferred API better.

veelo:
#636 adds the `qt` example, which looks like this: 
image

This `helloworld` example looks like this: 
image

I think it is OK to have both, but if I had to choose I'd vote for #636 because it uses the newest Qt binding.

PR #635, opened 2026-06-22
Abi changes coin5

VolkerEnderlein:


veelo:
Since the only data in `SbName` is a `const char*`, this is more about style than anything else, right? And with these changes there will be two pointer dereferences, whereas now there is just one -- or can the compiler optimise that out? Maybe it is worth comparing the differences with godbolt.org.

VolkerEnderlein:
Thanks for reviewing the changes. The changes were purely motivated by making the API more coherent, as most but not all methods get SbName's passed by reference. I checked with godbolt.org, and in Release mode the references are optimized away.

PR #638, opened 2026-07-03
tidbits: use C++ cmath functions

tritao:
## Summary
- Use `std::isinf` and `std::isnan` from `<cmath>` in `tidbits.cpp`.
- Preserve `coin_isinf()`'s documented `-1` / `+1` return contract with `std::signbit`.

## Motivation
Some C++ environments do not expose the unqualified C `isinf` and `isnan` names when including `<cmath>`. Coin detects these functions through the C `math.h` header, but `tidbits.cpp` is C++ code, so using the `std::` namespace is the portable form.

### DIFF
--- a/src/tidbits.cpp
+++ b/src/tidbits.cpp
@@ -1625,7 +1625,8 @@ int
 coin_isinf(double value)
 {
 #ifdef HAVE_ISINF
-  return isinf(value);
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks @tritao your ongoing contributions are highly appreciated.

PR #643, opened 2026-08-31
Fix SoCamera::getViewVolume viewport aspect ratio handling

tritao:
Fix `SoCamera::getViewVolume()` using the caller's previous `resultvp` contents when calculating the viewport aspect ratio.

## Problem

The method currently reads the aspect ratio from `resultvp` before assigning `vp` to it. This makes the returned view volume depend on the output parameter's initial state.

This affects public callers using:

```cpp
SbViewportRegion renderViewport;
SbViewVolume renderVolume =
  camera->getViewVolume(viewport, renderViewport);
```

## Fix

Calculate the aspect ratio directly from the input viewport:

```cpp
const float aspectratio = vp.getViewportAspectRatio();
```

The change is source- and ABI-compatible. Existing callers that initialize `resultvp` remain unaffected.

### DIFF
--- a/src/nodes/SoCamera.cpp
+++ b/src/nodes/SoCamera.cpp
@@ -434,7 +434,7 @@ SoCamera::getViewVolume(const SbViewportRegion & vp,
                         SbViewportRegion & resultvp, 
                         const SbMatrix & mm) const
 {
-  float aspectratio = resultvp.getViewportAspectRatio();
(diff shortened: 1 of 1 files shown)


VolkerEnderlein (approved):
LGTM, thanks for your work on the Coin codebase. It is highly appreciated.

tritao:
That was fast, thanks!

PR #645, opened 2026-09-01
Avoid null byte-buffer copies in `SbByteBuffer::push()`

tritao:
Guard zero-length memcpy calls in SbByteBuffer::push().

The existing pushOnEmpty test triggers UBSan because empty buffers expose null data pointers even though the copy size is zero. Avoid passing those pointers to memcpy.


VolkerEnderlein (approved):
LGTM, thanks for providing the fix.

PR #646, opened 2026-09-01
Establish profile-aware OpenGL headers

tritao:
This establishes a profile-aware OpenGL header and build-configuration foundation for Coin while preserving its public header contract and existing default renderer behavior.

## Motivation

Coin currently relies on platform OpenGL headers for both its public interface and internal implementation. This makes it difficult to move implementation code toward a defined core profile without also changing what installed consumers see.

This change separates the public compatibility path from a private core-profile path and separates renderer availability from the choice of GL declarations used by an individual source unit.

## Changes

- Vendor pristine Khronos core OpenGL and WGL headers for Coin’s private use, with exact upstream revisions recorded in `PROVENANCE.txt`.
- Extract the complete historical fallback block from `Inventor/system/gl.h` into `gl-fallbacks.h` without changing the public macro contract.
- Distinguish access to internal interfaces (`COIN_INTERNAL`) from compilation as part of the Coin library (`COIN_BUILDING_COIN`).
- Add an explicit private core-profile header that core-clean implementation units opt into directly, independently of renderer configuration.
- Reject mixing the private core path with desktop compatibility, `glext`, or GLES headers.
- Keep installed consumers on Coin’s public platform-header and compatibility-fallback path.
- Add `COIN_BUILD_LEGACY_GL_RENDERER`, enabled by default, and publish the resulting `COIN_HAVE_LEGACY_GL_RENDERER` capability.
- Keep Windows header policy such as `NOMINMAX` private to Coin’s build rather than defining it through an installed header.

## Compatibility

The default configuration continues to build and register the existing compatibility renderer. The vendored Khronos headers and the core-header selector remain private implementation details and do not become part of Coin’s public API.

With `COIN_BUILD_LEGACY_GL_RENDERER=OFF`, Coin omits legacy renderer registration and reports the capability as unavailable. Implementation objects still referenced elsewhere inside Coin remain part of the build; complete source-unit partitioning is follow-up work.

This was originally reviewed in https://github.com/FreeCAD/coin/pull/11


VolkerEnderlein:
@tritao this looks good to me. I have one question, is the switch from a self contained build and test YAML file to YAML+PWSH/SH script a solution that is commonly agreed on?

VolkerEnderlein:
Thank you @tritao for this amazing piece of work.

tritao:
> @tritao this looks good to me. I have one question, is the switch from a self contained build and test YAML file to YAML+PWSH/SH script a solution that is commonly agreed on?

The main idea behind it is to allow the same logic to be executed locally, for easier testing, then the CI workflow can be made simpler and just executes the same script.

PR #649, opened 2026-09-06
Don't require X11/GLX when cross-compiling for Android

Dikluwe:
## Summary

COIN_BUILD_GLX defaulted to ON unconditionally, and
`if(UNIX AND NOT APPLE AND COIN_BUILD_GLX)` treats Android as a match
(Android is UNIX-like, not APPLE) -- so a plain Android NDK
cross-compile would hit `find_package(X11 REQUIRED)` and fail the
configure step outright, even though Android has no X11/GLX at all,
only EGL (which COIN_BUILD_EGL already defaults on and is unaffected
by this change).

Made COIN_BUILD_GLX a cmake_dependent_option forced OFF when ANDROID
is set (the same idiom already used two lines above for
COIN_BUILD_MAC_X11/COIN_BUILD_MAC_AGL's APPLE-only defaults), rather
than requiring every Android build to remember to pass
-DCOIN_BUILD_GLX=OFF by hand.

Verified with `cmake -DANDROID=TRUE` (no real NDK toolchain available
in this environment, so this checks the CMake logic/configuration
path, not an actual Android cross-compile binary): the normal build
still shows COIN_BUILD_GLX=ON and populates 268 X11_* cache entries
from find_package(X11); with ANDROID=TRUE, COIN_BUILD_GLX is forced
off, find_package(X11) is never invoked (0 X11_* cache entries), and a
full `make Coin` completes with 0 errors using the EGL-only path
gl.cpp already supports (HAVE_EGL defined, HAVE_GLX undefined -- an
already-existing, working configuration in gl.cpp's own
HAVE_EGL-without-HAVE_GLX branch).

Also audited every other platform's offscreen-context-creation glue
function (gl_wgl.cpp for Windows, gl_cgl.cpp/gl_agl.cpp for macOS,
gl_glx.cpp for X11/Linux) for the same uninitialized-pointer-before-
assignment bug class as the one just fixed in gl_egl.cpp on a separate
branch: all four assign their context struct pointer before any use
of it, on every path -- clean, no fix needed. Since EGL is exactly
Android's OpenGL ES backend, the existing gl_egl.cpp fix already
covers what Android needs there; no separate Android-specific GL glue
work was required.

iOS was not touched: Coin has no glue at all for iOS's native context
API (EAGL is a distinct, Objective-C-only API neither AGL nor CGL
speak, and COIN_BUILD_MAC_AGL already explicitly excludes IOS), and
the existing APPLE-branch CMake logic falls through to probing for
CGL, which doesn't exist in the iOS SDK either -- getting proper iOS
support working is a real feature (writing a new EAGL-based glue
module), not a small config fix, so it's out of scope here.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,11 @@ cmake_dependent_option(COIN_BUILD_MAC_FRAMEWORK "Build framework instead of dyli
 cmake_dependent_option(COIN_BUILD_MAC_X11 "Build for X11 on Mac OS X when ON. Default is OFF." OFF "APPLE" OFF)
 cmake_dependent_option(COIN_BUILD_MAC_AGL "Build for AGL on Mac OS X when ON. Default is OFF." OFF "APPLE" OFF)
 
-option(COIN_BUILD_GLX "Build for GLX on Linux when ON. Default is ON." ON)
+# GLX needs X11, which Android does not have -- there is no windowing
+# system there for Coin to create an on-screen GLX context against,
+# only EGL. Force this off on Android rather than default it on and
+# fail later at find_package(X11 REQUIRED).
+cmake_dependent_option(COIN_BUILD_GLX "Build for GLX on Linux when ON. Default is ON." ON "NOT ANDROID" OFF)
 option(COIN_BUILD_EGL "Build for EGL on Linux when ON. Default is ON." ON)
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #652, opened 2026-09-06
Fix -Wcast-function-type in GLU NURBS begin-callback registration

Dikluwe:
## Summary

One site per NURBS shape class (SoNurbsCurve, SoNurbsSurface,
SoIndexedNurbsCurve, SoIndexedNurbsSurface): the GLU_NURBS_BEGIN_DATA
callback is registered by casting SoNurbsP<T>::tessBegin, whose real
signature is void(int, void*), straight to gluNurbsCallback_cb_t
(void(*)(void*, ...)). This is the standard GLU idiom -- the API
groups every NURBS callback selector (begin/vertex/normal/texcoord/
end/error, each with its own real signature) under one generic
registration type, and GLU internally calls back through the correct
concrete signature for the selector actually passed to
gluNurbsCallback(). The other 4 registrations in each of these files
(tessTexCoord/tessNormal/tessVertex/tessEnd) go through the identical
generic-registration idiom but don't trigger this warning, since only
a scalar (int) in a pointer-shaped slot -- as in tessBegin -- crosses
GCC's -Wcast-function-type threshold; the float*/void* mismatches on
the others don't.

Fixed with the standard, GCC/Clang-recognized suppression idiom for
intentional callback-dispatch casts: an intermediate cast through
void(*)(void) before the final cast to gluNurbsCallback_cb_t. This is
purely a type-system reinterpretation with no codegen effect -- cast
through void(*)(void) or not, the resulting function pointer is the
exact same value (confirmed with a standalone comparison: identical
address either way) -- so this changes nothing about how GLU actually
invokes the callback.

master except -Wcast-function-type 4->0 (859->855 total, zero errors);
full CoinTests suite passes (326 tests, 83566 checks) in both a normal
unrelated SbByteBufferP.icc UBSan finding present. As in the earlier
-Wunused-function pass over this same NURBS code, this sandbox's GLX/
EGL offscreen context creation doesn't work even for a trivial scene,
so the actual GLU tessellation callback invocation could not be
exercised at runtime here; the fix is a pure type-level reinterpret
with no behavioral effect on the call GLU makes, which is what the
above address-identity check demonstrates directly.

---

### DIFF
--- a/src/shapenodes/SoIndexedNurbsCurve.cpp
+++ b/src/shapenodes/SoIndexedNurbsCurve.cpp
@@ -345,7 +345,7 @@ SoIndexedNurbsCurveP::doNurbs(SoAction * action,
     this->nurbsrenderer = GLUWrapper()->gluNewNurbsRenderer();
 
     if (GLUWrapper()->versionMatchesAtLeast(1, 3, 0)) {
-      GLUWrapper()->gluNurbsCallback(this->nurbsrenderer, (GLenum) GLU_NURBS_BEGIN_DATA, (gluNurbsCallback_cb_t)SoNurbsP<SoIndexedNurbsCurve>::tessBegin);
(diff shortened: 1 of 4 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #654, opened 2026-09-06
Fix -Wclass-memaccess: replace memcpy over non-POD classes with assignment

Dikluwe:
## Summary

Three sites used memcpy() to copy objects of classes with user-declared
(or, for SoPointDetail, implicitly non-trivial due to a virtual
destructor) copy semantics, which GCC flags because memcpy is only
well-defined for TriviallyCopyable types -- these class(es) aren't
that, by definition, regardless of whether their current layout
happens to be safe to byte-copy:

- SoTextureCombineElement::UnitData::UnitData(const UnitData&)
  (src/elements/SoTextureCombineElement.cpp:317): memcpy(this, &org,
  sizeof(*this)) on a plain-data struct (all members are ints/enums/
  floats). Audited in an earlier -Wdeprecated-copy pass: this memcpy
  copies padding bytes too, while the class's own (now-defaulted)
  operator= copies member-by-member -- the two mechanisms only agree
  by coincidence of this struct's current all-4/8-byte-aligned layout
  having no padding. Replaced with a member-initializer-list copy
  matching the class's own default constructor's style, so copy ctor
  and operator= are provably doing the same thing by construction.

- soshape_primdata::shapeVertex()'s SoShape::POLYGON growth path
  (src/shapenodes/soshape_primdata.cpp:203,209): memcpy() over
  SoPrimitiveVertex[] and SoPointDetail[] arrays when growing them.
  SoPointDetail is polymorphic (inherits SoDetail, which declares
  virtual ~SoDetail() and pure virtual copy()/getTypeId()) --
  memcpy-ing over an already- constructed object of polymorphic type
  is undefined behavior per the standard even though, because both
  sides are always the concrete type SoPointDetail here, the vtable
  pointer bytes being copied are byte-identical and this doesn't
  corrupt anything in practice. Replaced both memcpy calls with
  per-element assignment loops using each class's own copy-assignment
  operator (hand-written and shallow-copy-by-design for
  SoPrimitiveVertex; implicit and trivial-per-member for
  SoPointDetail, whose SO_DETAIL_HEADER macro doesn't add any data
  members).

None of the 3 sites had a live, currently-observable bug: verified
with a full CoinTests run and a dedicated reproducer (an
SoIndexedFaceSet polygon face with 30 vertices, well past
soshape_primdata's initial array size of 4, forcing several grow()
calls) run under AddressSanitizer+UndefinedBehaviorSanitizer both
before and after this change -- clean in both cases. This is a
correctness-by-construction / portability fix (removes reliance on
implementation-defined behavior that happens to work on this
toolchain today), not a fix for an observed crash or corruption.

Added a regression test (src/shapenodes/SoIndexedFaceSet.cpp, under
COIN_TEST_SUITE) reproducing that same scenario through the public API
and checking the resulting triangulated geometry is correct, so a
future change that actually breaks copy-assignment for these classes
would fail on real data, not just re-add a compiler warning. Full
CoinTests suite (327 tests) passes clean under ASan+UBSan both before
and after.

---

Rule-of-Zero for SoTextureCombineElement::UnitData instead of hand-copy

Following up on the previous commit: rather than keep a hand-written
copy constructor that merely reproduces what the compiler would
generate anyway, remove it entirely. Every member of UnitData
(SbUniqueId, the enum arrays, SbColor4f, the floats) is itself
trivially copyable -- SbColor4f inherits SbVec4f, which is just a
`float vec[4]` with no user-declared special members -- so with no
user-declared copy ctor/dtor/operator= left, the compiler generates
all of them, and they are provably identical to what the hand-written
version did.

This is a strictly better fix than the initializer-list version from
the previous commit: it also makes UnitData trivially copyable
(verified with a static_assert), and, as a side effect, resolves the
-Wdeprecated-copy warning this class independently carried (it had a
user-declared copy ctor but no operator=, the same asymmetry fixed for
14 other classes on a separate branch) -- confirmed by full clean
rebuild: total warnings on this branch drop from 856 to 853, with
SoTextureCombineElement::UnitData no longer appearing in either the
-Wclass-memaccess or the -Wdeprecated-copy warning lists.

Full CoinTests suite (327 tests) still passes clean under ASan+UBSan.

---

Replace soshape_primdata's hand-rolled arrays with SbList<T>

Tried the deeper fix suggested for the remaining two -Wclass-memaccess
sites: instead of a manual new[]/assignment-loop/delete[] dance for
growing vertsArray/pointDetails, use SbList<SoPrimitiveVertex> and
SbList<SoPointDetail> (already used throughout Coin, and already
audited on another branch to grow via copy-assignment, never memcpy).

Mapped the access pattern carefully before doing this: fixed-size
shape types (triangle strip/fan, quads, quad strip, lines) only ever
touch indices 0-3, always within the 4 slots pre-populated in the
constructor; only SoShape::POLYGON grows past that, one vertex at a
time via growPolygonBuffers(), which mirrors exactly what the manual
code did -- append() one slot (SbList doubles its own buffer
internally when needed, same amortized cost as the old arraySize<<=1),
then re-point every vertsArray[i]'s SoPointDetail* at the (possibly
reallocated) pointDetails buffer. That re-pointing step is not
optional: SbList::append() reallocates just like the old code's
manual new[]/delete[] did, so previously-taken addresses go stale
exactly the same way in both versions -- this refactor does not change
that pre-existing pointer-stability contract, just who allocates.

getPointDetailIndex()'s pointer-arithmetic trick (recovering an index
from a previously-handed-out SoPrimitiveVertex*) now subtracts
vertsArray.getArrayPtr() instead of the old raw pointer -- same
contiguous-buffer assumption, still valid since SbList's storage is
contiguous too.

Validated before keeping this (was prepared to revert to the plain
memcpy->loop fix if anything showed a discrepancy):
- Full clean rebuild: 0 errors, 0 new warnings (853, same as the
  previous commit), 0 -Wclass-memaccess remaining.
- Full CoinTests suite (327 tests) passes under a plain build and
  under ASan+UBSan, with no new sanitizer findings (only the
  pre-existing, unrelated SbByteBufferP.icc null-argument report).
- The polygon_with_many_vertices_survives_array_growth regression test
  (added in the previous commit) still passes unchanged.
- A separate stress run with 500 polygon vertices (well past the
  original 30) under ASan+UBSan: no crash, and the resulting
  triangulation is geometrically correct (498 triangles, bbox exactly
  matching the input circle) -- growth through many reallocations
  doesn't corrupt data.

---


VolkerEnderlein:
Thanks for the contribution.

VolkerEnderlein:
@Dikluwe I think we need to revert the changes for the SoTextureCombineElement::UnitData. This breaks ABI compatibility with Coin 4.0.10. I checked that with a new ABI check action that runs on every PR. Can you please open a PR and fix the issue? Thanks in advance.

PR #658, opened 2026-09-06
Fix -Wformat= warnings (11 sites across 11 files)

Dikluwe:
## Summary

Two shapes:

- 10 sites (SoElement, SoEnvironmentElement, SoFloatElement,
  SoFontNameElement, SoInt32Element, SoLightAttenuationElement,
  SoReplacedElement, SoViewportRegionElement, SoState print()
  methods, plus mutex.cpp's lock-timing debug message): a `%p`
  conversion is given `this` (or, for mutex.cpp, a `cc_mutex *`)
  directly, but %p's argument must be exactly `void *` per the
  C standard -- passing any other pointer type is technically
  UB even though every real ABI represents all data pointers
  identically. Fixed with the standard, portable idiom: cast the
  argument to (void *) at the call site. Pure formatting-diagnostic
  code (print() methods, a debug log line), zero behavioral
  effect -- same bit pattern in, same characters out.

- 1 site (src/xml/element.cpp's cc_xml_elt_get_int32()): a real,
  if minor, correctness bug. This function was evidently copied
  from the cc_xml_elt_get_uint32() immediately above it (identical
  structure) but the format specifier was left as "%u" instead of
  being updated to "%d" for the now-signed int32_t * argument.
  Passing a signed target to a %u conversion is again technically
  UB, and can behave differently from %d for certain scanf() corner
  cases even where the representations coincide. Fixed by changing
  the specifier to "%d" to match the parameter's actual signedness.

master except -Wformat= 11->0 (859->848 total, zero errors); full
CoinTests suite passes (326 tests, 83566 checks) in both a normal
unrelated SbByteBufferP.icc UBSan finding present.

---

### DIFF
--- a/src/elements/SoElement.cpp
+++ b/src/elements/SoElement.cpp
@@ -595,7 +595,7 @@ void
 SoElement::print(FILE * file) const
 {
   (void)fprintf(file, "%s[%p]\n",
-                this->getTypeId().getName().getString(), this);
+                this->getTypeId().getName().getString(), (void *)this);
 }
 
 /*!
--- a/src/elements/SoEnvironmentElement.cpp
+++ b/src/elements/SoEnvironmentElement.cpp
@@ -261,7 +261,7 @@ SoEnvironmentElement::getFogType(SoState * const state)
 void
 SoEnvironmentElement::print(FILE * file) const
 {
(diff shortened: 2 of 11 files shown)


VolkerEnderlein:
Thanks for this contribution, too.

PR #659, opened 2026-09-06
Fix uninitialized ctx pointer dereference in eglglue_context_create_offscreen

Dikluwe:
## Summary

surface_attrib[] read ctx->width/ctx->height to build its EGL_WIDTH/
EGL_HEIGHT attributes, but ctx was an uninitialized local pointer at
that point -- it is only assigned a few lines later by
eglglue_contextdata_init(width, height). Every call dereferenced a
wild pointer before doing anything else.

Confirmed with a debugger-backed reproducer calling the real, exported
eglglue_context_create_offscreen() directly: it segfaults reliably
(SIGSEGV) at gl_egl.cpp:288 on real EGL/GPU hardware, independent of
Coin's GLX/EGL runtime auto-selection (this bypasses COIN_EGL and
gl.cpp's check_egl() entirely by calling the glue entry point
directly). Reaching this function at runtime requires a build with
HAVE_EGL (COIN_BUILD_EGL=ON, EGL dev libs found) and, when GLX is also
available, either COIN_BUILD_GLX=OFF (EGL becomes the only backend) or
the COIN_EGL=1 environment variable / an already-current EGL context
at runtime (gl.cpp check_egl()).

Fix: use the width/height function parameters directly instead of
ctx->width/ctx->height. eglglue_contextdata_init() sets
ctx->width = width and ctx->height = height a few lines down, so this
is provably the same value, not an invented default, and removes the
dependency on ctx for this array entirely.

Added a standalone reproducer (testsuite/reproducers/gl-egl-uninitialized-ctx/)
that calls the real glue function against a built libCoin.so and, via
an LD_PRELOAD shim intercepting eglCreatePlatformWindowSurface/
PixmapSurface, verifies the EGL_WIDTH/EGL_HEIGHT values actually
reaching EGL equal the requested size. Not wired into the default
ctest run since it needs real EGL headers/lib and ideally a real GPU,
which CI runners may lack; run manually via run.sh <build>/lib.

---

### DIFF
--- a/src/glue/gl_egl.cpp
+++ b/src/glue/gl_egl.cpp
@@ -282,11 +282,15 @@ eglglue_context_create_offscreen(unsigned int width, unsigned int height)
     EGL_NONE
   };
 
+  // Use the width/height arguments directly: ctx is not assigned until
+  // eglglue_contextdata_init() below, and eglglue_contextdata_init()
+  // sets ctx->width/ctx->height to exactly these values, so there is
+  // no need to (and, before ctx exists, no way to) read them off ctx.
   EGLAttrib surface_attrib[] = {
     EGL_TEXTURE_FORMAT, EGL_TEXTURE_RGBA,
     EGL_TEXTURE_TARGET, EGL_TEXTURE_2D,
-    EGL_WIDTH, (EGLint) ctx->width,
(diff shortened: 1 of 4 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #660, opened 2026-09-06
Fix -Wignored-qualifiers: drop meaningless top-level const on cast targets

Dikluwe:
## Summary

Both of the 34 warnings (8 distinct SoMFxxx::setValues() instantiations
plus 26 instantiations of the SO_SET_OVERRIDE macro) came from a single
macro-expansion site each:

- SoSubField.h's SO_MFIELD_VALUE_SOURCE macro:
  `static_cast<const _valtype_>(newvals[i])` casts a scalar value
  (SbBool/double/int/float/int32_t/short/uint32_t/unsigned short,
  depending on the field type) to a const-qualified type. A top-level
  const on the result type of a cast is meaningless -- the cast
  produces a prvalue, which was never assignable to begin with, so
  "const" changes nothing about it. Dropped the const; the conversion
  itself (and the value assigned into this->values[i+start]) is
  unchanged.

- SoOverrideElement.cpp's SO_SET_OVERRIDE macro:
  `const_cast<SoOverrideElement * const>(...)` -- same issue, one
  level of indirection over: this const-qualifies the pointer *value*
  produced by the cast (not what it points to), which is immediately
  discarded into `element`, itself already declared
  `SoOverrideElement * const` two lines up (that's the const that
  actually matters, on the named variable). Dropped the redundant
  const from the cast's target type.

No behavior change in either case -- these are the exact values that
were already being produced and assigned, just spelled without a
qualifier the compiler was already ignoring.

Full clean rebuild: 0 errors, 0 -Wignored-qualifiers remaining (was 34
across these 2 macro definitions), total warnings 859->825. Full
CoinTests suite (326 tests) passes.

---

### DIFF
--- a/include/Inventor/fields/SoSubField.h
+++ b/include/Inventor/fields/SoSubField.h
@@ -376,7 +376,7 @@ _class_::setValues(const int start, const int numarg, const _valtype_ * newvals)
   else if (start+numarg > this->num) this->num = start+numarg; \
  \
   for (int i=0; i < numarg; i++) \
-    this->values[i+start] = static_cast<const _valtype_>(newvals[i]); \
+    this->values[i+start] = static_cast<_valtype_>(newvals[i]); \
   this->setChangedIndices(start, numarg); \
   this->valueChanged(); \
   this->setChangedIndices(); \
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #664, opened 2026-09-06
Fix -Wmaybe-uninitialized warnings (41 sites, 2 root causes)

Dikluwe:
## Summary

39 of the 41 sites all trace back to the single 'obj' local in
SbHash<Key,Type>::operator[] (src/misc/SbHash.h:353): it declares
Type * obj uninitialized, tries getP(key,obj) and, on miss, does
put()+getP() again to fill it in, falling back to assert(false) if
that second getP still fails. Since assert() is compiled out under
NDEBUG, that fallback leaves obj uninitialized for the unconditional
`return *obj;` right after -- a real (if practically unreachable,
since a getP() immediately following a successful put() on the same
key should always succeed) uninitialized-pointer dereference in
release builds. Every instantiation of this template (SoBase's
name/object dictionaries, SoContextHandler, SoConfigSettings via
SbString::operator=, and SoGLDriverDatabaseP's 34 feature-lookup call
sites) inherited the same warning. Fixed at the root by initializing
`Type * obj = NULL;`, so the "should never happen" path now derefs a
null pointer -- a clean, diagnosable crash -- instead of reading
garbage.

The remaining 2 sites (SoShaderParameter.cpp:723, CGGLenum type/tform
in SoShaderStateMatrixParameter::updateParameter) are the same
assert-is-a-noop-in-release hazard: both are set inside a switch over
an SoSFEnum field, with only assert(0 && "illegal ...") in the
default case. Fixed by pre-initializing each to the field's own
documented default enum value (MODELVIEW / IDENTITY), mirroring how
the sibling updateValue() method in the same class already
pre-initializes its `matrix` local before an identical switch.

master except -Wmaybe-uninitialized 41->0 (859->818 total, zero

---

### DIFF
--- a/src/misc/SbHash.h
+++ b/src/misc/SbHash.h
@@ -350,7 +350,7 @@ class SbHash {
   }
 
   Type & operator[](const Key & key) {
-    Type * obj;
+    Type * obj = NULL;
     if (!getP(key,obj)) {
       Type dummy = Type();
       if (!put(key,dummy)) assert(false);
--- a/src/shaders/SoShaderParameter.cpp
+++ b/src/shaders/SoShaderParameter.cpp
@@ -700,7 +700,7 @@ SoShaderStateMatrixParameter::updateParameter(SoGLShaderObject *shader)
   this->ensureParameter(shader);
 
   if (shader->shaderType() == SoShader::CG_SHADER) {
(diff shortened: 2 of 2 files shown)


VolkerEnderlein:
Thanks.

PR #665, opened 2026-09-06
Fix S/T wrap-mode swap in SoMultiTextureImageElement/SoTexture2/SoTexture3

Dikluwe:
## Summary

SoMultiTextureImageElement::getWrapS()/getWrapT() returned each other's
storage slot (getWrapS() read unitdata[unit].wrapT and vice versa) ever
since the file was created in commit 48b16b44fb (2003, "build framework
for multi-texturing"), almost certainly copy-pasted from the older
single-texture SoTextureImageElement predecessor.

Independently, SoTexture2::doAction() and both call sites in
SoTexture3::doAction() passed wrapS/wrapT to
SoMultiTextureImageElement::set() in swapped order too. This traces back
further: the pre-2005 single-texture-only code path in SoTexture2 already
called SoTextureImageElement::set() with wrapT and wrapS backwards
relative to that element's own declared (wrapS, wrapT, ...) signature.
Commit 4d3ab98b1e (2005, "Support for multiple texture units also for
SoCallbackAction") copy-pasted that already-swapped argument order into
the new SoMultiTextureImageElement::set() call added for unit != 0, and
commit 665411f3e1 (2009) later unified the two code paths, making the
swapped call the only path, used for all texture units including 0.

Because the getter swap and the setter-call-site swap are both
S<->T transpositions, they canceled out for 25 years on the one path that
exercises both: SoCallbackAction/SoRayPickAction introspection via
doAction() -> SoMultiTextureImageElement::set() -> later read back through
SoCallbackAction::getTextureWrapS()/getTextureWrapT(). That's why the bug
was never observed in practice, even though wrong values were stored
internally the whole time.

SoTexture2::GLRender() and SoTexture3::GLRender() were never affected:
they call SoGLImage::setData() directly with correctly-ordered
translateWrap() calls, bypassing SoMultiTextureImageElement storage
entirely, so actual OpenGL rendering was always correct.

Also checked and confirmed unaffected: SoImage.cpp (passes the same
CLAMP literal for both S and T), and vrml97/ImageTexture.cpp /
PixelTexture.cpp (independently authored, correctly map repeatS/repeatT
to the set() call's wrapS/wrapT slots).

Fixing only one side of the swap (getter or setter) would have broken
the previously-"working" (accidentally correct) SoCallbackAction path;
both had to be fixed together. Verified end-to-end with a standalone
SoCallbackAction-based smoke test exercising SoTexture2 with distinct
wrapS/wrapT values through a real (non-default) image.

Verified: full GCC + clang builds with -Wall -Wextra -Wpedantic
-Wno-write-strings show no warning-count change (862 total, matching
current master baseline -- these three files contribute exactly one
pre-existing, unrelated -Wdeprecated-copy warning). CoinTests passes
1/1 via ctest on both compilers. Debug+ASan/UBSan run reports
[OK] 326 tests, 83566 checks, zero UBSan findings, and no new leaks
(all pre-existing leak reports trace to unrelated SoSeparator
construction in miscSoBaseTest.cpp).

---

### DIFF
--- a/src/elements/SoMultiTextureImageElement.cpp
+++ b/src/elements/SoMultiTextureImageElement.cpp
@@ -452,7 +452,7 @@ SoMultiTextureImageElement::getWrapS(SoState * const state, const int unit)
     (getConstElement(state, classStackIndex));
 
   PRIVATE(elem)->ensureCapacity(unit);
-  return PRIVATE(elem)->unitdata[unit].wrapT;
+  return PRIVATE(elem)->unitdata[unit].wrapS;
 }
 
 /*!
@@ -466,7 +466,7 @@ SoMultiTextureImageElement::getWrapT(SoState * const state, const int unit)
     (getConstElement(state, classStackIndex));
 
   PRIVATE(elem)->ensureCapacity(unit);
-  return PRIVATE(elem)->unitdata[unit].wrapS;
+  return PRIVATE(elem)->unitdata[unit].wrapT;
 }
 
 /*!
--- a/src/nodes/SoTexture2.cpp
+++ b/src/nodes/SoTexture2.cpp
@@ -643,8 +643,8 @@ SoTexture2::doAction(SoAction * action)
   if (size != SbVec2s(0,0)) {
     SoMultiTextureImageElement::set(state, this, unit,
                                     size, nc, bytes,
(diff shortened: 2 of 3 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #666, opened 2026-09-06
Fix GCC -Wnull-dereference warnings across the codebase (COIN_ASSUME + isOfType+cast pattern)

Dikluwe:
Harden missing-element read paths and checked node casts against NULL before dereferencing, retaining the existing valid-input behavior. Include focused reproducers for disabled elements and indexed-shape failure cases. The affected drivers now finish Coin only after their local objects have been destroyed.

## Validation


VolkerEnderlein:
Thanks for the contribution.

PR #669, opened 2026-09-06
Fix -Warray-bounds false positive in SbList<Type>::grow()

Dikluwe:
## Summary

GCC flagged 8 call sites across the codebase (SoOutput::reset,
SoReorganizeActionP::initShape, SoNodekitCatalog::reallyAddEntry,
SoBaseKit::createFieldList, SbHeap::emptyHeap, SoLightPath::setHead,
Background.cpp's buildGeometry) with "array subscript 0 is outside
array bounds of 'Type* [0]'" / "note: object of size 0 allocated by
operator new []", pointing at SbList.h:248, the element-copy loop in
grow().

Audited every mutator of itembuffersize (constructor sets it to
DEFAULTSIZE; grow() only ever doubles it or raises it to an explicit
size > the current one; fit() clamps it to max(items, DEFAULTSIZE))
and confirmed by full-suite runs under AddressSanitizer/UBSan (before
and after) that itembuffersize can never actually be 0 when grow()
runs new Type[itembuffersize] -- there is no real out-of-bounds write.

Bisected the false positive to a minimal, Coin-independent repro: it
requires code that sets numitems to 0 (e.g. truncate(0), or a
getLength()==0 guard) immediately before an append(), fully inlined
together. All 8 real call sites match this exact shape (6 via
truncate(0)+append(), one via an empty-the-list loop + append(), one
via if(getLength()==0){append();}). GCC constant-propagates
numitems==0 into the inlined grow(), and -- unable to prove
itembuffersize can't also be 0 across the constructor/inlining
boundary -- reports the (actually unreachable) "new Type[0]" branch.

Fix: give grow() a floor that is provable from purely local
information, instead of relying on cross-function knowledge GCC
doesn't have here: `itembuffersize > 0 ? itembuffersize << 1 :
DEFAULTSIZE` instead of a bare `itembuffersize <<= 1`. Confirmed this
eliminates the warning (0 remaining after a full clean rebuild, was 8)
with the exact same object code behavior for the only state that ever
actually occurs (itembuffersize >= DEFAULTSIZE); no pragma or
suppression involved.

Added a regression test (src/lists/SbList.cpp, under COIN_TEST_SUITE)
that exercises the real truncate(0)-then-grow-past-DEFAULTSIZE pattern
end to end and checks the resulting values, so a future change that
actually breaks the itembuffersize invariant would fail a real memory
check, not just silence a compiler warning. Full CoinTests suite
(327 tests after this addition) passes clean under ASan+UBSan both
before and after the fix.

---

### DIFF
--- a/include/Inventor/lists/SbList.h
+++ b/include/Inventor/lists/SbList.h
@@ -238,8 +238,17 @@ protected:
 
 private:
   void grow(const int size = -1) {
-    // Default behavior is to double array size.
-    if (size == -1) this->itembuffersize <<= 1;
+    // Default behavior is to double array size. itembuffersize is
+    // always >= DEFAULTSIZE for the lifetime of the object (set in
+    // the constructor, and every other assignment below either
+    // doubles it or clamps it to at least DEFAULTSIZE), so the ">
+    // 0" branch is the only one ever taken in practice -- but the
+    // explicit floor lets the compiler prove that locally too,
+    // instead of just at this call's actual call sites, which is
+    // what silences a GCC -Warray-bounds false positive seen when
+    // append() is inlined right after code that resets numitems to
+    // 0 (e.g. truncate(0)) without narrowing itembuffersize.
+    if (size == -1) this->itembuffersize = (this->itembuffersize > 0) ? (this->itembuffersize << 1) : DEFAULTSIZE;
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #670, opened 2026-09-06
Fix -Wshift-negative-value in SpiderMonkey glue's INT_TO_JSVAL

Dikluwe:
## Summary

All 37 warnings are the same macro, INT_TO_JSVAL(i), expanded at every
call site (JSVAL_VOID's own definition, plus every val = JSVAL_VOID;
in JS_VRMLClasses.cpp/Script.cpp). jsval is intptr_t (signed), and
JSVAL_VOID's argument (0 - JSVAL_INT_POW2(30)) is negative, so
`(jsval)(i) << 1` left-shifts a negative signed integer -- undefined
behavior pre-C++20, and still warned about by GCC regardless of
-std level.

This header hand-reimplements SpiderMonkey's own jsapi.h bit-packing
scheme for Coin's dlopen-based glue, so the computed value has to stay
byte-for-byte identical to what the real library expects; this is a
warning-only fix, not a value change. Fixed by doing the shift in
uintptr_t (well-defined for any bit pattern, including the sign bit)
and casting the result back to jsval, which reproduces the exact same
bit pattern as the original left-shift on any two's-complement target
(i.e. every platform Coin actually runs on) -- confirmed with a
standalone comparison of both macro forms across JSVAL_VOID's actual
argument plus a spread of boundary values (0, +-1, +-2^30, INT32_MIN/
MAX): identical output in every case.

master except -Wshift-negative-value 37->0 (859->822 total, zero

---

### DIFF
--- a/include/Inventor/C/glue/spidermonkey.h
+++ b/include/Inventor/C/glue/spidermonkey.h
@@ -215,7 +215,7 @@ struct JSErrorReport {
 #define JSVAL_INT_MAX           (JSVAL_INT_POW2(30) - 1)
 #define INT_FITS_IN_JSVAL(i)    ((uint32_t)((i)+JSVAL_INT_MAX) <= 2*JSVAL_INT_MAX)
 #define JSVAL_TO_INT(v)         ((int32_t)(v) >> 1)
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #671, opened 2026-09-06
Fix -Wsign-compare: heap overflow in cc_xml_load_file() on ftell() error

Dikluwe:
## Summary

`cc_xml_load_file()` used `ftell()` without checking for failure. On a
non-seekable stream such as a FIFO, `ftell()` returns `-1`; the subsequent
signed/unsigned conversions could make `fread()` request `SIZE_MAX` bytes into
a zero-byte allocation.

Check the `ftell()` result before allocating, return `NULL` on failure, and use
`size_t` only after the non-negative value has been validated. Also change the
unrelated non-negative size threshold in
`SoInput_FileInfo::readUnsignedIntegerString()` from `int` to `size_t`, removing
the second `-Wsign-compare` occurrence without changing behavior.

The POSIX reproducer compiles the private `src/xml/utils.cpp` implementation
directly into its executable. It does not link Coin and does not require
`cc_xml_load_file()` to be exported from the Coin DLL. It now requires a
`NULL` result when `ftell()` fails instead of treating any non-crashing return
as success.

## Validation

- Before the fix, the instrumented FIFO reproducer aborts with
  `buffer overflow detected`.
- After the fix, the reproducer reports the required `NULL` result under both
  GCC and Clang with `-Wall -Wextra -Wpedantic -Wsign-compare` and
  ASan/UBSan enabled for the implementation and driver.
- Full Release builds and `ctest` pass with GCC and Clang.
- A Debug ASan/UBSan build instruments Coin and `CoinTests`; all 326 tests and
  83,566 checks complete. The default sanitizer run is not clean because it
  reports a pre-existing null-pointer argument in `SbByteBufferP.icc` and
  11,552 leaked bytes in 207 allocations from unrelated tests. With leak
  detection disabled, `ctest` exits successfully; the UBSan report remains a
  validation gap outside this PR.

Static inspection confirmed that `cc_xml_load_file()` has no caller in the
Coin codebase and remains unexported. The reproducer is manual and POSIX-only;
Windows/MSVC and macOS were not rerun for this revision.


src/xml/utils.h
```diff
@@ -41,7 +41,7 @@ extern "C" {
 
 /* ********************************************************************** */
 
-char * cc_xml_load_file(const char * path);
+COIN_DLL_API char * cc_xml_load_file(const char * path);
```
VolkerEnderlein: No, this should not be exported from the Coin DLL in any case. It is **not** used over the entire Coin codebase so the fix itself is valid, but the reproducer should not require the DLL to export unused functions.

VolkerEnderlein (changes requested):

src/xml/utils.h
Dikluwe: Agreed, fixed in 8f0d2899 — dropped `COIN_DLL_API` from `cc_xml_load_file()` in utils.h and changed the reproducer to compile `src/xml/utils.cpp` directly instead of linking against libCoin, so the function stays private. `run.sh` now builds it standalone with ASan/UBSan.

VolkerEnderlein:
Thanks for the contribution.

PR #672, opened 2026-09-06
Fix precision/type mismatches found while auditing -Wuseless-cast

Dikluwe:
## Summary

Neither fix relates to useless-cast itself; both were found by reading
the full context around flagged lines while triaging that category.

src/base/SbDPRotation.cpp: SbDPRotation::slerp() used FLT_EPSILON as
its near-antipodal-rotation threshold, copy-pasted unchanged from
SbRotation.cpp's float-precision slerp() since the file's original
2002 commit (473266a656). In the narrow band where (1-dot) falls
between DBL_EPSILON and FLT_EPSILON, this silently fell back to
linear interpolation instead of proper spherical interpolation,
discarding the extra precision a caller chose SbDPRotation for.
Switched both threshold checks to DBL_EPSILON (already available via
the file's existing <cfloat> include).

src/engines/SoConvertAll.cpp: the SOCONVERTALL_CAST_SFIELD2FIELD
macro table's SoSFBool_SoSFUInt32/SoSFBool_SoMFUInt32 entries used
`unsigned short` as the intermediate cast type, while every other
conversion to SoSFUInt32/SoMFUInt32 in the same table uses uint32_t.
Comparison against the SoSFBool_SoSFUShort/SoSFBool_SoMFUShort
entries just below (which correctly use unsigned short, for a
different 16-bit target) suggests these were copy-pasted without
updating the type. Currently harmless in practice -- the source is
always SoSFBool (0 or 1), which round-trips losslessly either way --
but a real type inconsistency in the field-conversion machinery.

---


VolkerEnderlein:
Thank you for the contribution.

PR #673, opened 2026-09-06
Deprecate SoJavaScriptEngine, make its availability check trustworthy

Dikluwe:
## Summary

SoJavaScriptEngine binds to the SpiderMonkey C API as it existed
around 2007 (JSBool, JS_GetPrivate()/JS_SetPrivate(), no JS::Rooted<>
requirement). Git history shows ~2 years of active API-tracking on
this glue (2005-2007), then essentially nothing but copyright-header
and build-system touches for the ~17 years since -- coinciding with
SpiderMonkey's embedding API going through multiple incompatible
rewrites that nobody kept chasing, and with VRML97 itself (the only
caller of this engine, via SoVRMLScript's "javascript:"/"vrmlscript:"
URLs) having been superseded by X3D since 2004.

This is public API (include/Inventor/misc/SoJavaScriptEngine.h), not
unreachable dead code, so it isn't a delete-outright candidate the way
an internal, uncalled function would be -- removing a public class is
an API break reserved for a major version bump. Followed this
project's own existing pattern instead (see 456aefc834, "Deprecate
SoInput::readHex()": add a \deprecated Doxygen tag, change nothing
about behavior). Documented on SoJavaScriptEngine's class comment, and
cross-referenced from SoVRMLScript (the node applications actually
touch) so its docs explain why "javascript:" URLs silently do nothing
on a current system rather than looking like an oversight.

Separately, found and fixed a real safety gap while investigating this
(confirmed against the actual mozjs-115 package now installed in this
environment): spidermonkey()'s per-symbol REGISTER_FUNC() macro only
assert()s on a missing symbol -- a no-op in release builds -- while
sm->available starts optimistically TRUE and is only ever reset to
FALSE if the library fails to *open*. If a library opens successfully
but doesn't export the expected flat C symbols (exactly what happens
with any current SpiderMonkey, which exports a name-mangled C++ API
instead), available stays TRUE over a set of NULL function pointers.
Callers that correctly check spidermonkey()->available before using
the API (Script.cpp, SoJavaScriptEngine.cpp both do) would be misled
into calling through those NULL pointers.

Verified end-to-end with the real libmozjs-115 package now installed:
  - master (unfixed), COIN_SPIDERMONKEY_LIBNAME=libmozjs-115.so.0:
    dlopen succeeds, available=1, JS_NewRuntime/JS_NewContext/
    JS_EvaluateScript all NULL -- the exact landmine described above.
  - this branch, same environment: dlopen succeeds, available is
    correctly reset to 0, with a debug-mode message identifying why
    ("Found a SpiderMonkey library, but it does not export the
    (obsolete) API this glue expects").
Fix re-derives availability after all REGISTER_FUNC calls from whether
the handful of functions needed to create a runtime, create a context,
and evaluate a script actually resolved, rather than trusting the
initial optimistic default.

No plan to port this glue to a current SpiderMonkey release as part of
this: its embedding API keeps changing enough that such a port isn't a
one-time cost, and there's no evidence of anyone depending on it (zero
automated test coverage of SoJavaScriptEngine or JS_VRMLClasses in
testsuite/). That's a separate, much larger decision for whoever owns
this project's roadmap, not something to fold into a warning/hygiene
pass.

compile errors and identical warning counts to the master baseline
(862, same per-category breakdown), with no new warnings at any
touched line. CoinTests passes 1/1 via ctest on both compilers.
Debug+ASan/UBSan run reports [OK] 326 tests, 83566 checks, zero UBSan
findings, and the same 207 pre-existing leak allocations as the
established baseline.

---

### DIFF
--- a/src/glue/spidermonkey.cpp
+++ b/src/glue/spidermonkey.cpp
@@ -192,17 +192,48 @@ spidermonkey(void)
 
 
   /* Define macro for grabbing function symbols. Casting the type is
-     necessary for this file to be compatible with C++ compilers. */
+     necessary for this file to be compatible with C++ compilers.
+
+     A missing symbol clears sm->available and logs a debug-mode
+     message rather than assert()ing: a library that opens successfully
+     but doesn't export a given flat C symbol (e.g. any current
+     SpiderMonkey, which exports a name-mangled C++ API instead of the
+     ~2007-era C API this glue expects) is an ordinary, expected runtime
+     condition -- "found a library, but not the one this glue can
+     drive" -- not a programming-error invariant violation, so it must
+     be handled the same way in Debug and Release builds instead of
+     aborting the whole process via a failed assert() only in Debug.
+     Tracking this per-symbol (rather than re-deriving availability
+     afterward from a hand-picked subset of "the important" functions)
+     also means every symbol registered below counts, automatically,
+     including ones added here in the future -- not just whichever
+     handful someone remembered to list in a separate check. */
   #define REGISTER_FUNC(_funcname_, _funcsig_) \
           sm->_funcname_ = (_funcsig_)cc_dl_sym(spidermonkey_libhandle, SO__QUOTE(_funcname_)); \
-          assert(sm->_funcname_)
+          if (!sm->_funcname_) { \
+            sm->available = 0; \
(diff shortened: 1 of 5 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #676, opened 2026-09-06
Sync SFNodeEnginePath.tpl/MFNodeEnginePath.tpl isOfType+cast pattern

Dikluwe:
## Summary

These two .tpl files are the historical source templates that produced
SoSFNode/SoSFPath/SoSFEngine/SoMFNode/SoMFPath/SoMFEngine.cpp via
scripts/templant (each generated .cpp still carries the
"//$ BEGIN/END TEMPLATE" markers and a header comment saying it "was in
full generated ... from the code in [the .tpl]"). templant is not wired
into the CMake build and hasn't been for a long time, so this sync has
no effect on compiled output -- it's purely keeping the documented
source of truth from silently reverting future re-generation, if anyone
ever ran templant again, back past 2008's introduction of
coin_assert_cast (the .tpl still had the original 1999-era plain
C-style casts) and now past this branch's coin_safe_cast +
if (T* x = ...) re-expression fix for -Wnull-dereference.

Scope, deliberately narrow: only the writeValue()/write1Value() and
countWriteRefs() isOfType+cast pattern, plus the "SbBasicP.h" include
it needs, matching exactly what was already fixed in the six generated
.cpp files two commits back on fix/null-dereference.

Explicitly NOT touched, left as documented pre-existing drift: the .tpl
has diverged from the generated files in several other, unrelated ways
that predate this work by years -- most notably, SoSFNode.cpp's
readValue() has a VRML97-specific NULL-child special case
(isFileVRML1()/isFileVRML2() handling) that exists ONLY in that one
file, not in the other five, and was never templated at all. Folding
that (or the other scattered C-style-to-coin_assert_cast conversions in
setValue()/checkCopy()) back into the shared .tpl would risk silently
changing behavior for types it was never written for. That's a
separate, much larger reconciliation project than a warning fix.

Verified: since .tpl files aren't part of the build, there's no
warning-count delta possible from this change. Confirmed by diffing the
updated .tpl (with its NODE/Node/node placeholders substituted) against
the actual fixed SoSFNode.cpp/SoMFNode.cpp content that the synced
blocks now match exactly, modulo the deliberately-untouched drift noted
above. Ran a plain default-flags build + ctest as a basic sanity check
(1/1 pass) since the six actual .cpp files were already fully verified
on the parent commit.

---

### DIFF
--- a/src/fields/MFNodeEnginePath.tpl
+++ b/src/fields/MFNodeEnginePath.tpl
@@ -22,6 +22,9 @@
 #include <Inventor/fields/SoMF_Typename_.h>
 #include <Inventor/fields/SoSubFieldP.h>
 #include <Inventor/fields/SoSF_Typename_.h>
+
+#include "SbBasicP.h"
+
 #include <Inventor/SoOutput.h>
 #include <Inventor/actions/SoWriteAction.h>
 #include <Inventor/SoPath.h>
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #677, opened 2026-09-06
Fix -Wundef debug-diagnostic bugs in SoField and SoVRMLGroup

Dikluwe:
Use the runtime coin_debug_extra() switch for SoField diagnostics and restore the missing GL-cache debug toggle in SoVRMLGroup. Access container information safely in the container-less field diagnostic. The regression driver now destroys its local field/engine objects before Coin shutdown.

## Validation

### DIFF
--- a/src/base/SbDPMatrix.cpp
+++ b/src/base/SbDPMatrix.cpp
@@ -332,7 +332,7 @@ double
 SbDPMatrix::det3(int r1, int r2, int r3,
                int c1, int c2, int c3) const
 {
-#if COIN_EXTRA_DEBUG
+#ifdef COIN_EXTRA_DEBUG
   // Check indices.
   if (r1<0 || r1>3 || r2<0 || r2>3 || r3<0 || r3>3 ||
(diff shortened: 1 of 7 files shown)


VolkerEnderlein (changes requested):
The previously `COIN_DEBUG_EXTRA` guarded sections should be changed to a common pattern that is used throughout the coin code base:

```
#if COIN_DEBUG
  if (coin_debug_extra()) {
  }
#endif // COIN_DEBUG
```
or
```
#if COIN_DEBUG
  if (coin_debug_normalize()) {
  }
#endif // COIN_DEBUG
```

This is currently not enforced but in a lot of other places it is already done this way.
This means the release code does not contain debugging specific code. Only the debug version contains this code and it can be enabled by setting the specific environment variable.

Dikluwe:
The requested release-build exclusion is now applied: every affected `coin_debug_extra()` diagnostic block is enclosed in `#if COIN_DEBUG` / `#endif`, so the debug-only code is not compiled into release builds. The branch also includes the follow-up cleanup requested during review. Could you please take another look?

PR #678, opened 2026-09-06
Fix -Wunused-but-set-variable warnings (54 sites across 27 files)

Dikluwe:
## Summary

Three shapes, by far the dominant one being the same assert-is-a-noop-
in-release hazard already fixed for -Wunused-variable elsewhere in
this codebase (see 74204c497b): a side-effecting call's result is
assigned to a local and only ever checked via assert(), which
compiles out under NDEBUG, leaving the local genuinely unread in a
release build:

  - src/threads/{mutex,condvar,thread}.cpp (13 sites): every
    lock/unlock/init/clean/wake/wait wrapper around the platform
    thread primitives. Fixed by turning the trailing assert(ok [==
    CC_OK]) into an if/assert (mutex.cpp keeps `ok`, since it's
    assigned from different platform-#ifdef branches; condvar.cpp/
    thread.cpp collapse to `if (call() != CC_OK) assert(!"...")`
    since each has a single assignment site).
  - src/vrml97/JS_VRMLClasses.cpp (1 template site, 10 warnings from
    10 instantiations), src/misc/{SoDB,SoProto,SoJavaScriptEngine,
    SoDBP}.cpp, src/fields/SoFieldContainer.cpp (3 sites),
    src/fonts/{freetype,glyph}.cpp, src/glue/freetype.cpp,
    src/tidbits.cpp: same pattern, one call site each. tidbits.cpp
    and SoDBP.cpp needed care since their calls (setlocale(),
    coin_vsnprintf()) have real side effects that must still execute
    in release builds -- fixed by keeping the call as a plain
    statement and only wrapping the diagnostic comparison in the
    if/assert, never the call itself.
  - src/base/rbptree.cpp and src/rendering/SoGLNurbs.cpp: same
    consumption pattern but the right-hand side is a pure, side-
    effect-free expression (`&rbptree_sentinel`, `FindSpan(...)`), so
    here the call/expression was inlined directly into the assert
    instead, which is safe specifically because nothing depends on it
    running in release builds.

Second shape: purely dead output of an otherwise-relevant computation,
with no assert consuming it either -- SoFieldContainer.cpp's
`sftype`, SoGlyph.cpp's `max`, SoRenderManager.cpp's `oldorigin`,
SoAsciiText.cpp's `originalmaxx`, SoCamera.cpp's `oldvp`,
SoShadowGroup.cpp's `smoothing` (explicitly dead per its own "FIXME:
temporary until we have time to fix this feature" comment),
SoExtSelection.cpp's `vpo`/`vps` (and the `vp` they were the only
consumers of), VectorizeActionP.cpp's `vpsize` (and its sole producer
`vp`), SoNodeVisualize.cpp's `point`, SbPolarStereographic.cpp's
`phiF` (present and used in the sibling project() method, genuinely
unused in unproject()), SoVertexAttribute.cpp's `setvbo`, and 8
version-number locals in SoGLDriverDatabase.cpp::findDriver() feeding
a driver-matching feature its own comment says isn't implemented yet.
All of these are pure getters/literals with no side effect, so deleted
outright.

Third shape, a real bug found while investigating rather than a
hygiene issue: VectorizeActionP.cpp's shade_vertex() computes an
alpha value `A = SbClamp(vcolor[3], 0.0f, 1.0f)` for every vertex but
its return statement only ever passed 3 arguments to SbColor4f(r,g,b,
a=1.0f), silently defaulting alpha to fully opaque and discarding the
input vertex's actual transparency. Fixed by passing A through to the
constructor's 4th argument -- this is the one behavioral change in
this branch; every other file here is a pure warning fix with no
change to release-build behavior.

master except -Wunused-but-set-variable 54->0 (859->805 total, zero

---

### DIFF
--- a/src/base/rbptree.cpp
+++ b/src/base/rbptree.cpp
@@ -266,14 +266,12 @@ static void
 rbptree_delete_fixup(cc_rbptree * t, cc_rbptree_node * x)
 {
   /* page 274 */
-  cc_rbptree_node * w, * nil;
-
-  nil = &rbptree_sentinel;
+  cc_rbptree_node * w;
 
   while (x != t->root && x->color == RBPTREE_BLACK) {
     if (x == x->parent->left) { /* x is left child */
       w = x->parent->right;
-      assert(w != nil);
+      assert(w != &rbptree_sentinel);
       if (w->color == RBPTREE_RED) {
         w->color = RBPTREE_BLACK;
         x->parent->color = RBPTREE_RED;
@@ -300,7 +298,7 @@ rbptree_delete_fixup(cc_rbptree * t, cc_rbptree_node * x)
     }
     else { /* x is right child */
       w = x->parent->left;
-      assert(w != nil);
+      assert(w != &rbptree_sentinel);
       if (w->color == RBPTREE_RED) {
         w->color = RBPTREE_BLACK;
         x->parent->color = RBPTREE_RED;
--- a/src/fields/SoFieldContainer.cpp
+++ b/src/fields/SoFieldContainer.cpp
@@ -876,10 +876,8 @@ SoFieldContainer::addCopy(const SoFieldContainer * orig,
   assert(copiedinstances);
   assert(contentscopied);
 
-  SbBool s = copiedinstances->put(orig, copy);
-  assert(s);
-  s = contentscopied->put(orig, FALSE);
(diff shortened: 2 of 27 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #679, opened 2026-09-06
Fix clang unused-but-set warnings and STL buffer cleanup

Dikluwe:
Preserve and check the accumulated STL read/write status in Debug builds, return the binary facet writer's status, remove unused bookkeeping, and add binary/ASCII round-trip coverage. This remains stacked on the GCC unused-but-set-variable cleanup.

Release the previous ASCII header buffer before rescanning after the format probe, clean it on rejection, and free writer->filename on destruction. Keep steel.l and the checked-in generated steel.cpp synchronized. The /dev/full reproducer now cleans up the parent process; the child's intentional abort/_exit is not a leak-clean shutdown check.

## Validation


VolkerEnderlein:
Thanks for the contribution.

PR #680, opened 2026-09-06
Fix clang -Wunused-const-variable in SoConfigSettings.cpp

Dikluwe:
## Summary

The `COIN` byte-string constant in the anonymous namespace was only
ever referenced from inside SoConfigSettings::reinitialize()'s
environ-scanning loop, which is entirely wrapped in
`#if COIN_DEBUG && 0` -- permanently preprocessed out (per the
existing FIXME: "environ is not available on non-unix platforms, so
disabling this for now"), so the declaration had no live reference in
any actually-compiled translation unit.

Moved the declaration inside the same #if COIN_DEBUG && 0 block, right
before its only uses, instead of deleting it -- deleting it would
break the disabled code if COIN_DEBUG_ENVIRON scanning is ever
resurrected once the cross-platform environ availability is sorted
out, same reasoning as this audit's -Wconstant-logical-operand fixes
for the other deliberately-disabled-but-kept debug blocks in this
codebase. Zero behavior change: the block was never compiled before
this fix and still isn't.

Verified: -Wunused-const-variable 1 -> 0 under clang -Wall -Wextra
-Wpedantic (390 -> 389 warnings total), 0 errors, every other category
in the breakdown unchanged. GCC build unaffected (859 warnings, 0
errors, same as master baseline). Full CoinTests suite passes under
both compilers (normal build: 1/1 ctest pass each); GCC Debug+ASan/
UBSan: 326 tests, 83566 checks, no findings.

---

### DIFF
--- a/src/misc/SoConfigSettings.cpp
+++ b/src/misc/SoConfigSettings.cpp
@@ -18,7 +18,6 @@ namespace {
   {
     return sizeof(VALID_OPTIONS)/sizeof(VALID_OPTIONS[0]);
   }
-  const char COIN [] = "COIN";
 
   SbBool isValidOption(const SbString & option)
   {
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #681, opened 2026-09-06
Fix unused-function warnings and remove dead helpers

Dikluwe:
## Summary

Resolve unused-function warnings by matching debug and optional-feature guards to their callers, annotating intentionally retained helpers, and removing genuinely dead XML and NURBS test code. The standalone NURBS surface-point evaluator remains available; it is marked intentionally unused because Coin currently delegates discretisation to GLU.

This PR no longer changes GL context ownership or destruction. The earlier cc_glglue deallocation experiment was reverted and remains isolated in a study branch because post-destruction pointer usage requires a separate lifetime contract.

## Current diff

- 9 files changed
- 50 insertions and 148 deletions
- no change to src/glue/gl.cpp relative to master

## Validation

- The Coin library itself also builds with unused-function promoted to an error. Pre-existing unused helper warnings in generated test translation units prevent applying that Werror policy to the entire testsuite.

The sanitizer run also observes a pre-existing SoReorganizeAction downcast diagnostic in the STL round-trip test when UBSan is configured to abort on the first finding. It is outside this PR and is recorded as a validation limitation rather than attributed to this change.

Windows, macOS and optional-feature combinations were not rerun locally; CI remains authoritative for those configurations.

### DIFF
--- a/src/rendering/SoGLNurbs.cpp
+++ b/src/rendering/SoGLNurbs.cpp
@@ -452,18 +452,13 @@ namespace {
       delete [] Bin[0];
       delete [] Bin;
     }
-    void Test();
-
     void BinomialCoefficients(const int rows, const int cols);
     int FindSpan(const float u, const int degree, const int numctrlpts, const float* knotvec) const;
-    void BasisFunctions(const float u, const int span, const float* knotvec, const int order, float* N);
     void DersBasisFunctions(const float u, const int span, const float* knotvec, const int order, const int n, float** ders );
     void RationalSurfaceDerivsH(const float u, const float v, const int d, SbVec4f** skl);
     void RationalSurfaceDerivs(const float u, const float v, const int d, SbVec3f** skl);
     SbVec3f normal(const float u, const float v);
     SbVec3f normal(const int i, const int j);
-    SbVec4f SurfacePoint(const float u, const float v);
-    SbVec4f SurfacePoint(const int i, const int j);
     void mapijtouv(const int i, const int j, float& u, float &v);
 
   private:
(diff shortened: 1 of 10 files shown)


src/rendering/SoGLNurbs.cpp
```diff
@@ -766,161 +744,6 @@ namespace {
     v = (j == numvctrlpts-1) ? vknotvec[numvknots-1] : (vknotvec[0] + j*deltav);
   }
 
-  /// Returns the point on the surface at parameters (u,v).
-  SbVec4f nurbs::SurfacePoint(const float u, const float v) {
```
veelo, line 770: I have mixed feelings about removing `SurfacePoint`. The reason for it not being used appears to be that Coin defers discretisation to GLU. But it seems incomplete to me to have functions for derivatives but not the points themselves. That's typically the first thing you want from a NURBS implementation.

I can't say that this is important to me, if I am not mistaken SARC B.V. have custom discretisation procedures. I would have to go back in time to dig up the reason for that, but it could be because the GLU discretisation did not meet requirements. If that is the case, these functions may prove themselves valuable to someone.

Dikluwe: Good point. I restored both `SurfacePoint` overloads and their `BasisFunctions` helper, while leaving the file-writing `Test()` routine removed. The declarations use this PR’s `COIN_UNUSED_FUNC` marker so the preserved evaluator does not reintroduce `-Wunused-function`. I validated commit 155f8121a5 with a full GCC 13 Release Coin build using `-Werror=unused-function`; all 906 compile/link steps passed.

PR #682, opened 2026-09-06
Fix -Wunused-parameter warnings (67 sites across 52 files)

Dikluwe:
## Summary

All are parameters read only inside code that a normal (NDEBUG)
release build compiles out: assert()-only consumption, #if COIN_DEBUG
blocks, platform #ifdef branches that don't apply here (e.g.
cc_internal_ftime() on Linux, qmeshGetWeight() without HAVE_ILOGB), or
stub/deprecated functions whose bodies were never implemented
(SoGeo::toGD/fromGD, SoRenderManager::attachClipSensor,
stl_facet_set/get_padding). Fixed with the existing COIN_UNUSED_ARG()
idiom (src/coindefs.h), reusing it in files that already use it and
adding the "coindefs.h" include where missing.

The largest single cluster (25 sites) is every Sb*::print(FILE * fp)
debug-dump method across src/base/ -- the whole body is guarded by
#if COIN_DEBUG, so fp is provably unused in a release build. Care was
taken not to add the include inside an existing #if COIN_DEBUG block
(would make the macro itself invisible in release builds) and not to
duplicate it where a file already pulled it in under a different
spelling (<coindefs.h> vs "coindefs.h").

steel.l/steel.cpp are Flex source and its checked-in generated output
(included verbatim via steel-wrapper.cpp, with #line directives
pointing back to steel.l, which is why the warnings report as
steel.l); both were edited in lockstep to keep them byte-identical.

master except -Wunused-parameter 67->0 (859->792 total, zero errors);
full CoinTests suite passes (326 tests, 83566 checks) in both a normal
unrelated SbByteBufferP.icc UBSan finding present.

---

### DIFF
--- a/src/3ds/3dsLoader.cpp
+++ b/src/3ds/3dsLoader.cpp
@@ -1799,7 +1799,7 @@ SoTextureCoordinate2* FaceGroup::createSoTextureCoordinate2_n(tagContext *con)
 
 
 
-SoTriangleStripSet* FaceGroup::createSoTriStripSet_n(tagContext *con)
+SoTriangleStripSet* FaceGroup::createSoTriStripSet_n(tagContext * COIN_UNUSED_ARG(con))
 {
   assert(!con->useIndexedTriSet && "Improper use.");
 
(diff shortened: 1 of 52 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #683, opened 2026-09-06
Fix clang -Wunused-private-field across 22 sites

Dikluwe:
## Summary

Clang reports `-Wunused-private-field` at 22 sites. Most are private pointers
reserved in exported classes so future implementation data can be added
without changing the class layout. Removing those fields would break ABI.

Initialize the reserved fields in constructor bodies and keep their public
header declarations unchanged. With Clang 18, an initializer-list entry such
as `pimpl(nullptr)` still triggers `-Wunused-private-field`, while an
assignment in the constructor body is recognized as a use. This avoids all
compiler-specific diagnostic pragmas in public headers.

Three element constructors were previously generated by `SO_ELEMENT_SOURCE`.
They now use `SO_ELEMENT_CUSTOM_CONSTRUCTOR_SOURCE` and explicitly preserve
the macro's type-id and stack-index initialization before assigning their
reserved pointer. Four private implementation back-pointers similarly move
their existing initializer-list assignment into the constructor body.

`SoProtoInstance::classfielddata` remains present and is initialized to
`NULL`; it is superseded by `pimpl->fielddata`, but retaining it preserves the
layout of the exported class. `SoStream::reallocCallback` and its private
typedef are removed because they are internal, never assigned, and never
read.

## Validation

- A Clang 18 Release build of the base revision fails with
  `-Werror=unused-private-field`, first at `SoNormalBundle::pimpl`.
- The final Clang Release build completes with
  `-Wall -Wextra -Wpedantic -Werror=unused-private-field`; therefore no
  `-Wunused-private-field` occurrence remains.
- Full GCC and Clang Release builds and `ctest` pass.
- `sizeof` was compared for all 17 affected public classes between the base
  revision and the final revision; every size is identical.
- A GCC Debug ASan/UBSan build instruments Coin and `CoinTests`; all 326 tests
  and 83,566 checks complete. The default run is not a clean leak check because
  LeakSanitizer reports 11,552 bytes in 207 allocations from unrelated suite
  tests. With leak detection disabled, `ctest` passes.

The affected declarations, constructors, macro-generated constructors,
private back-pointers, and excluded/internal code were inspected. No pragma or
`COIN_UNUSED_ARG(master)` suppression remains. Windows/MSVC and macOS were not
rerun for this revision.

### DIFF
--- a/include/Inventor/VRMLnodes/SoVRMLAnchor.h
+++ b/include/Inventor/VRMLnodes/SoVRMLAnchor.h
@@ -71,14 +71,7 @@ private:
   static SoVRMLAnchorCB * fetchurlcb;
   static void * userdata;
 
-#ifdef __clang__
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunused-private-field"
-#endif
   SoVRMLAnchorP * pimpl;
-#ifdef __clang__
-#pragma clang diagnostic pop
-#endif
 };
 
 #endif // ! COIN_SOVRMLANCHOR_H
--- a/include/Inventor/VRMLnodes/SoVRMLAnchor.h
+++ b/include/Inventor/VRMLnodes/SoVRMLAnchor.h
@@ -71,7 +71,14 @@ private:
   static SoVRMLAnchorCB * fetchurlcb;
   static void * userdata;
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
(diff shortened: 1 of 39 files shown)


VolkerEnderlein (changes requested):
Are these warnings still triggered if the private fields are assigned a value (e.g. nullptr) in the constructor? If not, then I would prefer that solution instead of adding walls of preprocessor code to get rid of this warning.

VolkerEnderlein:
Thanks for the contribution.

PR #684, opened 2026-09-06
Fix -Wunused-variable: assert()/COIN_DEBUG-only consumption pattern

Dikluwe:
## Summary

136 of the 181 -Wunused-variable warnings (102x 'ok', 32x 'quadok', 2x
'isok') came down to just 19 distinct source lines, all sharing one
shape: a boolean/count is captured from a call that has an essential
side effect (populating an out-parameter, or removing/inserting a hash
entry), and the variable is then read only inside assert(...) or an
#if COIN_DEBUG block -- both compiled out entirely in a release
(NDEBUG) build, leaving the variable genuinely unused there while the
call it was assigned from still needs to run for its side effect.

Followed the exact idiom the actual Coin maintainer already
established for this identical problem in SbHash::operator[]
(74204c497b, "Preserve SbHash::operator[] behavior while avoiding a
status variable that becomes unused when assertions are disabled"):
replace `Type ok = call(...); assert(ok);` with
`if (!call(...)) assert(false);` (or `assert(!"message")` for the
`assert(ok && "message")` variant seen in JS_VRMLClasses.cpp) -- the
call keeps running unconditionally in every build, and the boolean is
consumed by the `if` in all builds too, not just debug ones.

SoQuadMesh.cpp's single flagged line (32 warnings, one per template
instantiation of SoGL::QuadMesh::GLRender) needed the same idea
adapted to its #if COIN_DEBUG shape: moved the `if (!qmeshNormalize(...))`
out of the #if so the call and the check both always run, and left
only the SoDebugError::postWarning() call -- the actually
debug-only part -- inside #if COIN_DEBUG.

One exception left untouched: JS_VRMLClasses.cpp:484-493 has a
similarly-named `ok` that IS reassigned and read across several
statements in a loop, which puts it under the separate
-Wunused-but-set-variable diagnostic instead (a different, still-open
warning category, not touched here).

No behavior change: every affected call still executes exactly when
it did before, in every build configuration; only how its result is
consumed changed. Full clean rebuild: 0 errors, 136 fewer
-Wunused-variable warnings (181->45 remaining, all differently-named,
one-off variables not part of this pattern), total warnings 859->723.
Full CoinTests suite (326 tests) passes under a plain build and under
ASan+UBSan, with no new sanitizer findings.

---

### DIFF
--- a/src/fields/SoField.cpp
+++ b/src/fields/SoField.cpp
@@ -230,8 +230,7 @@ public:
 
   void removeConverter(const void * item)
   {
-    size_t ok = this->maptoconverter.erase(item);
-    assert(ok);
+    if (!this->maptoconverter.erase(item)) assert(false);
   }
 
   SoFieldConverter * findConverter(const void * item)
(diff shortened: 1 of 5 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #685, opened 2026-09-06
Fix remaining unused-variable warnings

Dikluwe:
## Summary

Resolve the remaining one-off unused-variable warnings that were intentionally kept separate from the larger assert-only migration.

The final diff against current master touches 31 files. It preserves calls with required side effects, moves pure declarations into the compile-time regions that consume them, and removes declarations that are genuinely dead. The branch was merged with current master without changing the already validated source tree.

## Validation

- The Coin library itself builds with unused-variable promoted to an error. Pre-existing unused-variable warnings in test sources prevent applying that Werror policy to the entire testsuite.

The sanitizer run also observes a pre-existing SoReorganizeAction downcast diagnostic in the STL round-trip test when UBSan is configured to abort on the first finding. It is outside this PR and is recorded as a validation limitation rather than attributed to this change.

Windows, macOS and optional-feature combinations are covered by CI rather than this local rerun.

### DIFF
--- a/src/base/SbBSPTree.cpp
+++ b/src/base/SbBSPTree.cpp
@@ -270,7 +270,13 @@ coin_bspnode::split()
 
   this->dimension = dim; // set the dimension
 
-  float mid = (box.getMin()[dim] + box.getMax()[dim]) / 2.0f;
+  // Also read from in the #if COIN_DEBUG block further down (which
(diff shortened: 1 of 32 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #687, opened 2026-09-06
Add explicit casts for provably-bounded narrowing in gzmemio.cpp / SoKeyboardEvent.cpp

Dikluwe:
gzmemio.cpp (all 4 sites, the entirety of this tree's C4267): each
assigns/subtracts the size_t return of cc_gzm_fread() into a uint32_t
or unsigned int, but every call site already bounds that return value
to fit: two are capped by the constant Z_BUFSIZE, one by avail_out
(itself an unsigned int), and the last is only assigned after a
"remain > nmemb" check proves nmemb smaller than a uint32_t already.
No behavior change; casts document why each is safe.

SoKeyboardEvent.cpp: build_convert_dicts() computes a lowercase/
uppercase ASCII letter as 'a'/'A' + int arithmetic, but the loop it's
in only ever runs for i in [A, Z], so the result always fits in a
char. Cast documents that bound.

C4267 drops from 4 to 0, C4244 drops by 2 more, 0 errors, full
CoinTests suite passes (326 tests, 83566 checks).

---

### DIFF
--- a/src/events/SoKeyboardEvent.cpp
+++ b/src/events/SoKeyboardEvent.cpp
@@ -163,7 +163,8 @@ build_convert_dicts(void)
        i++
        )
     {
-      d->put(i, ('a' + i - static_cast<int>(SoKeyboardEvent::A)));
+      /* i ranges over A..Z here, so this is always a lowercase ASCII letter. */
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #689, opened 2026-09-06
Fix a real size_t underflow and suppress two intentional-dead-code warnings

Dikluwe:
SoFieldContainer.cpp: getFieldsMemorySize()'s "unsupported field type"
fallback set elementsize (a size_t) to -1, which wraps to SIZE_MAX
rather than signaling "unknown" the way it would for a signed type;
that value then flows straight into "managed/unmanaged += elementsize
* numelements" with no check in between, corrupting this diagnostic
API's memory-usage total for any field of a type this function
doesn't recognize. Fix: contribute 0 instead, matching elementsize's
own already-zero declaration default -- an accurate "we don't know
this type's size" rather than a value that overflows the total.

SoGeo.cpp / systemsanity.icc: find_utm_projection() and
SoDB_compileTimeAsserts() are both intentionally-unused (kept for
future use / exist only to fail compilation if their static asserts
don't hold) and already marked with this codebase's COIN_UNUSED_FUNC
macro -- but that macro is __attribute__((__unused__)) on GCC and
empty everywhere else (src/coindefs.h), so it does nothing to silence
MSVC's C4505 for the same functions. Suppress locally with
#pragma warning(push/disable(4505)/pop) around each, rather than
extend COIN_UNUSED_FUNC itself, since its GNU-attribute placement
(trailing the declaration) isn't where an MSVC pragma can go.

0 errors, full CoinTests suite passes (326 tests, 83566 checks).

---

### DIFF
--- a/src/fields/SoFieldContainer.cpp
+++ b/src/fields/SoFieldContainer.cpp
@@ -1243,8 +1243,12 @@ SoFieldContainer::getFieldsMemorySize(size_t & managed, size_t & unmanaged) cons
       } else if (mftypekey == MFVec4us_string.getString()) {
         elementsize = sizeof(SbTypeInfo<SoMFVec4us>::DataType[2]) / 2;
       } else {
-        // unsupported field type
-        elementsize = -1;
+        // unsupported field type -- contribute nothing to the total
+        // rather than corrupting it: elementsize is a size_t, so the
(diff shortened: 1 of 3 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #690, opened 2026-09-06
Fix MSVC /W4 false positive in SoTexture3::loadFilenames

Dikluwe:
volumenc is set on the first image loaded per call (guarded by
this->images.isDefault()) and read on every subsequent one in the
same call, but the compiler can't see the two-part guarantee that
makes this safe across repeat calls: this->images.setValue() clears
the default flag as a side effect (so images 1..N-1 correctly take
the "verify against volumenc" branch within one call), while a
successful call ends by explicitly setting the flag back to TRUE (the
existing "write filenames, not images" comment a few lines down) --
so isDefault() is true again at the start of the next call, and the
next call's first image correctly takes the "set volumenc" branch
again rather than reading a stale value from the previous call.

Initialized only because the compiler can't correlate isDefault()
with whether volumenc was actually set; does not change behavior on
this already-safe path.

this was the single remaining C4701 warning in the tree (after the
four preceding /W4 branches), 0 errors, full CoinTests suite passes
(332 tests, 83688 checks, tested merged into lab/all-fixes-integration).

---

### DIFF
--- a/src/nodes/SoTexture3.cpp
+++ b/src/nodes/SoTexture3.cpp
@@ -457,7 +457,16 @@ SoTexture3::loadFilenames(SoInput * in)
 {
   SbBool retval = FALSE;
   SbVec3s volumeSize(0,0,0);
-  int volumenc;
+  /* Only read below once this->images.isDefault() is false, which only
+     happens after this->images.setValue() below has run (setValue()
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #691, opened 2026-09-06
Clean up dead test-code MSVC /W4 flags in embedded test suites

Dikluwe:
- SbDPPlane.cpp: delX, delY and count in equalityToFloatPlane are
  declared and never read anywhere in the test case; dead leftovers.
- ScXMLMinimumEvaluator.cpp: res and boolobj are only used by code
  that's entirely commented out in this test case, which is itself
  marked "not finished" (FIXME from 2009); the two declarations are
  dead until that code is written.
- SbString.cpp: createInstance() returns a fixed fake pointer and is
  never called from anywhere in this file; dead leftover.
- SoBaseP.cpp: readok is only referenced inside assert(readok), which
  the preprocessor drops entirely under NDEBUG (Release builds),
  making it genuinely unused there. Same fix already established
  elsewhere in this codebase for this exact pattern (see mutex.cpp):
  move the read into an if() so it survives NDEBUG.

C4189 drops from 68 to 62 and C4505 from 21 to 20 (all 6 targeted
sites gone, no new warnings), 0 errors, full CoinTests suite passes
(326 tests, 83566 checks).

---

### DIFF
--- a/src/base/SbDPPlane.cpp
+++ b/src/base/SbDPPlane.cpp
@@ -438,9 +438,6 @@ BOOST_AUTO_TEST_CASE(signCorrect)
 
 BOOST_AUTO_TEST_CASE(equalityToFloatPlane)
 {
-  const float delX = 1;
-  const float delY = .1f;
-
   const float XMax = (float)pow(2.,FLT_MAX_EXP/3.);
   const float XMin = -XMax;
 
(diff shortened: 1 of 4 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #692, opened 2026-09-06
Guard SoQuadMesh's unimplemented 4D-coordinate weight calculation

Dikluwe:
The high-quality rendering path computes per-vertex weights (w1..w4)
from squared distances (s1..s4), but s1..s4 are only assigned on the
is3d branch. The else branch (4D/homogeneous coordinates) has all of
its equivalent assignments commented out, with a comment marking the
whole 4D path "not currently implemented" and a matching
assert(!"4d coordinates handling unimplemented yet") a few lines
above -- which, like any assert, is compiled out under NDEBUG
(Release builds), so s1..s4 would silently read indeterminate values
there instead of hitting that assert.

MSVC's /W4 flags this as C4701 across several instantiations of the
surrounding rendering-path code. Since the 4D path is a pre-existing,
explicitly-documented gap rather than something to implement here,
fix by zero-initializing s1..s4, so the already-broken 4D path reads
deterministic zeros instead of stack garbage.

C4701 drops from 97 to 17 (the surrounding code is instantiated
multiple times via macro for different rendering configurations, so
this one fix clears many sites at once), 0 errors, full CoinTests
suite passes (326 tests, 83566 checks).

---

### DIFF
--- a/src/shapenodes/SoQuadMesh.cpp
+++ b/src/shapenodes/SoQuadMesh.cpp
@@ -479,7 +479,13 @@ namespace { namespace SoGL { namespace QuadMesh {
       SbVec4f ccd4;
       SbVec4f sum234d4,sum134d4,sum124d4,sum123d4;
       SbVec4f vec1d4,vec2d4,vec3d4,vec4d4;
-      float s1,s2,s3,s4;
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #693, opened 2026-09-06
Fix MSVC /W4 warnings in Windows-only GDI/WGL/dlopen code

Dikluwe:
These four files are only ever compiled on Windows (guarded by
HAVE_WIN32_API / WGL / HAVE_WINDLL_RUNTIME_BINDING), so the Linux/GCC
warning-cleanup pass this session had no way to reach them -- these
warnings surfaced only now, building on Windows with /W4.

win32.cpp: flww32_getVerticesFromPath() uses lastmoveto as an array
index into p_points on a PT_CLOSEFIGURE, but only sets it on a prior
PT_MOVETO. A GDI path always starts with PT_MOVETO per the Win32 docs,
so this can't happen in practice, but an uninitialized index feeding
an array access is worth guarding regardless of how the path was
built; initialized to 0, the same index a leading PT_MOVETO would set.

gl_wgl.cpp: pixformat is always set by wglChoosePixelFormat() in the
same loop iteration that later makes context->hpbuffer non-NULL, but
the compiler can't correlate that across the loop's several
continue/break paths. False positive; initialized to 0.

time.cpp / dl.cpp: BOOL b / DWORD retval are each only referenced
inside their following assert(), which the preprocessor drops
entirely under NDEBUG (Release builds), making the variable
genuinely unused there. Same fix already established elsewhere in
this codebase for this exact pattern (see mutex.cpp): move the read
into an if() so it survives NDEBUG, then assert(!"message") on
failure.

C4701 drops from 97 to 95 and C4189 from 68 to 66 (all 4 targeted
sites gone, no new warnings), 0 errors, full CoinTests suite passes
(326 tests, 83566 checks).

---

### DIFF
--- a/src/base/time.cpp
+++ b/src/base/time.cpp
@@ -95,7 +95,7 @@ cc_internal_queryperformancecounter(cc_time * COIN_UNUSED_ARG(t))
   if (highperf_available) {
     LARGE_INTEGER counter;
     BOOL b = QueryPerformanceCounter(&counter);
-    assert(b && "QueryPerformanceCounter() failed even though QueryPerformanceFrequency() worked");
+    if (!b) assert(!"QueryPerformanceCounter() failed even though QueryPerformanceFrequency() worked");
     *t = (double)counter.QuadPart * highperf_tick + highperf_start;
     return TRUE;
   }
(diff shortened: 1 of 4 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #694, opened 2026-09-06
Fix MSVC /W4 potentially-uninitialized false positives

Dikluwe:
MSVC's C4701/C4703 flag several variables the compiler can't prove
are always set before use, even though the surrounding control flow
guarantees it in every case: short-circuited conditions that both
assign and test a variable in the same expression (SoNodeVisualize.cpp,
SoWriterefCounter.cpp), a ternary that conditionally calls an
out-param function (SoGLImage.cpp x2, SoGLBigImage.cpp), a value only
read under the same guard that sets it (SoImage.cpp, SoInput.cpp), a
call to an external out-param function the compiler can't see into
(SoSFNode.cpp), and an invariant enforced by a separate grammar file
rather than local control flow (evaluator.c).

Fix each by initializing to the value the surrounding logic already
implies: SoWriterefCounter's own "not found" sentinel, SoSFNode's own
explicit-NULL branch value, GL's own default pixel zoom, zero
components/chars/counts where that's the safe empty-case default.
This does not change behavior on any of these (already-safe) paths;
SoGLBigImage.cpp and SoGLImage.cpp's createGLDisplayList also harden
a narrower path where getImage()/this->image could plausibly be NULL,
turning an indeterminate allocation size into a defined zero.

C4701 drops from 97 to 85 (all 9 targeted sites gone, no new
warnings), 0 errors, full CoinTests suite passes (326 tests, 83566
checks).

---

### DIFF
--- a/src/engines/evaluator.c
+++ b/src/engines/evaluator.c
@@ -185,7 +185,12 @@ dot_product(float *v0, float *v1)
 void
 so_eval_traverse(so_eval_node *node, so_eval_param *result, const so_eval_cbdata *cbdata)
 {
-  so_eval_param param1, param2, param3;
+  /* Each param is only read below by opcodes whose arity the grammar
+     (evaluator.y) guarantees matches the number of children set on
+     the node, so e.g. a binary opcode's node always has child1 and
+     child2. Zero-initialized only because the compiler can't see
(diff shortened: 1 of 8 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #695, opened 2026-09-06
Fix testsuite floatEquals() sign bug found via -Wtype-limits

Dikluwe:
floatEquals() (testsuite/TestSuiteMisc.h), the ULP-based float
comparison helper used throughout the test suite via
COIN_TESTCASE_CHECK_FLOAT-style checks, reinterprets a float's raw
bits through a union to lexicographically order it for comparison --
the classic technique requires reading those bits as a *signed*
int32_t so the "< 0" check can detect the sign bit and remap negative
floats into the same ordered sequence as positive ones. This union
declared the field as uint32_t instead, so "if (A.i32 < 0)" was always
false (exactly what GCC's -Wtype-limits was flagging, 300 times across
every generated test file that includes this header) and the remap
never ran.

Concretely: -0.0f and 0.0f are numerically equal but have raw bit
patterns 0x80000000 and 0x00000000 -- ~2^31 apart. With the remap
disabled, floatEquals(-0.0f, 0.0f, N) incorrectly reports "not equal"
for any small N; more generally, any two floats straddling zero (e.g.
-1e-40f vs 1e-40f) suffer the same false-negative. Same-sign
comparisons away from the zero boundary happen to still work, since
the raw bit ordering is internally consistent there, which is likely
why this went unnoticed.

Fix: declare the union's integer field as int32_t, restoring the
sign-bit check and remap. Added a regression test
(floatEquals_negativeZero in testsuite/TestSuiteMisc.cpp) that fails
under the old code and passes under the fix.

Verified on Linux (GCC and clang): -Wtype-limits drops from 300 to 0
following a required CMake reconfigure (testsuite/CMakeLists.txt's
create_testsuite() macro extracts each src/*.cpp file's embedded
COIN_TEST_SUITE block into its own generated test source at configure
time, so a plain rebuild alone won't pick this up -- `cmake` needs to
be re-run once), full CoinTests suite passes (334 tests) under a plain
build and under ASan+UBSan, with no new sanitizer findings.

---

### DIFF
--- a/testsuite/TestSuiteMisc.cpp
+++ b/testsuite/TestSuiteMisc.cpp
@@ -88,3 +88,22 @@ operator << (std::ostream & os, const SbColor4f & col4)
   os << col4[3];
   return os;
 }
+
+BOOST_AUTO_TEST_CASE(floatEquals_negativeZero)
+{
+  // -0.0f and 0.0f are numerically equal but have different raw bit
+  // patterns (sign bit set vs. clear). floatEquals()'s ULP comparison
+  // (above in TestSuiteMisc.h) relies on remapping negative floats'
+  // raw bits into the same lexicographic order as positive ones,
+  // which requires reading those bits through a *signed* integer type
+  // so the "< 0" check can detect the sign bit at all -- with an
+  // unsigned type that check can never trigger, no remapping happens,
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #696, opened 2026-09-06
Fix clang -Wmissing-braces in heap.cpp's embedded tests

Dikluwe:
mock_up::wrapped_value is a one-member struct ({ double x; }), so
initializing an array of it with a flat list ("val[] = {3, 2, 1, ...}")
elides the per-element braces clang's -Wmissing-braces wants ("suggest
braces around initialization of subobject") -- GCC accepts this form
silently under -Wall/-Wextra, clang doesn't. Same 7-element literal is
repeated identically across all 5 test cases in this block (35 sites
total).

Fix: brace each element ("{{3}, {2}, {1}, ...}"), which both compilers
accept identically -- no behavior change, this is pure aggregate
initialization syntax.

Verified on Linux (GCC and clang): -Wmissing-braces drops from 35 to 0
on clang (GCC was already silent here), full CoinTests suite passes
under a plain build and under ASan+UBSan, with no new sanitizer
findings.

---

### DIFF
--- a/src/base/heap.cpp
+++ b/src/base/heap.cpp
@@ -407,7 +407,7 @@ public:
 };
 
 BOOST_AUTO_TEST_CASE(min_heap) {
-  mock_up::wrapped_value val[] = {3, 2, 1, 15, 5, 4, 45};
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #698, opened 2026-09-06
Fix -Wunused-parameter across testsuite callback signatures

Dikluwe:
11 sites across 5 files, all the same shape: a callback registered
via a *Callback(func, userdata)-style API where the function doesn't
need the userdata (or, in one case, the error object) it's handed.
Each function's signature is fixed by the callback typedef it must
match, so the type stays -- only the now-unused parameter name is
dropped, matching the standard C/C++ idiom for this.

- TestSuiteMain.cpp: main()'s argc/argv (CoinTest::run_all() takes no
  arguments and reads neither).
- TestSuiteUtils.cpp: the data (userdata) parameter of all three
  Coin error-handler callbacks (debug/read/memory) -- each already
  gets what it needs from the error object itself.
- SoTransformerDragger.cpp: the SoCallbackAction parameter of
  register_cb() (SbDict/SoNode are all it needs), and the value
  parameter of ensure_unique_cb() (shadowed by its own local `val`
  a line below -- the two were never the same parameter).
- SoMFNode.cpp / SoDB.cpp: an identical do-nothing readErrorHandler()
  (both error and data unused) duplicated verbatim in each file's own
  embedded test block.

Verified on Linux (GCC and clang): -Wunused-parameter drops from 11 to
0 on both, full CoinTests suite passes under a plain build and under
ASan+UBSan, with no new sanitizer findings.

---

### DIFF
--- a/src/draggers/SoTransformerDragger.cpp
+++ b/src/draggers/SoTransformerDragger.cpp
@@ -2219,7 +2219,7 @@ SoTransformerDragger::setDynamicRotatorSwitches(const SoEvent *event)
 
 static
 SoCallbackAction::Response
-register_cb(void * data, SoCallbackAction * action, const SoNode * node)
+register_cb(void * data, SoCallbackAction *, const SoNode * node)
 {
   assert(data);
   SbDict * dict = static_cast<SbDict *>(data);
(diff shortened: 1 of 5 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #699, opened 2026-09-06
Fix unused test locals and conditional scene-graph cleanup

Dikluwe:
Remove or conditionally compile unused variables in the embedded tests for SbDPRotation, SoType, SoSFVec4us, ScXMLMinimumEvaluator and SoBase, preserving required side effects.

In checkWriteWithMultiref, release the original graph outside HAVE_VRML97: it is allocated even when the conditional checks are excluded. The converted VRML graph remains cleaned up inside that block.

## Validation

### DIFF
--- a/src/fields/SoSFVec4us.cpp
+++ b/src/fields/SoSFVec4us.cpp
@@ -131,9 +131,8 @@ BOOST_AUTO_TEST_CASE(textinput)
   SoSFVec4us field;
   field.set("1 2 3 4");
   BOOST_CHECK_EQUAL(field.getValue(), SbVec4us(1, 2, 3, 4));
-  const char * filters[] = { "read error", NULL }; // all read error messages
   TestSuite::ResetReadErrorCount();
-  // TestSuite::PushMessageSuppressFilters(filters);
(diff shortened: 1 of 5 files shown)


src/fields/SoSFVec4us.cpp
```diff
@@ -131,9 +131,8 @@ BOOST_AUTO_TEST_CASE(textinput)
   SoSFVec4us field;
   field.set("1 2 3 4");
   BOOST_CHECK_EQUAL(field.getValue(), SbVec4us(1, 2, 3, 4));
-  const char * filters[] = { "read error", NULL }; // all read error messages
   TestSuite::ResetReadErrorCount();
-  // TestSuite::PushMessageSuppressFilters(filters);
+  // TestSuite::PushMessageSuppressFilters(filters); // filters = { "read error", NULL }
```
VolkerEnderlein: I would not do this change. Rather than adding the `filters` variable after the function call, I would comment out the entire `const char* filters[]` line and leave the rest unchanged.

VolkerEnderlein (changes requested):

VolkerEnderlein:
Thanks for the contribution.

PR #702, opened 2026-09-06
Fix SoTextureCubeMap distortion on SoCube (issue #563)

### LINKED ISSUE #563: SoTextureCubeMap incorrect UV mapping
SoTextureCubeMap seems to map the textures incorrectly. I tried setting the `filenames` field with 6 images and also the `image*` fields individually but in both cases the textures look distorted. Texture mapping a SoCube with a single SoTexture2 seems to work fine.

With SoTextureCubeMap:
Image

With SoTexture2:
Image

Here is some code for reference:
```cpp
auto get_cubemap_filenames() {
	const static char *filenames[] = {
		"right.jpg", "left.jpg",
		"top.jpg", "bottom.jpg",
		"front.jpg", "back.jpg"
	};

	return filenames;
}

void add_cube(SoSeparator *root) {
	auto cube_separator = new SoSeparator;
	auto ts = new SoTranslation;
	ts->translation = {2, 0, 0};
	cube_separator->addChild(ts);

	SoInput::addDirectoryFirst("../assets/skybox");

	// SoTexture2 *texture = new SoTexture2;
	// texture->filename = "back.jpg";
	// cube_separator->addChild(texture);

	auto cube_map = new SoTextureCubeMap;
	cube_map->filenames.setValues(0, 6, get_cubemap_filenames());
	cube_separator->addChild(cube_map);

	auto cube = new SoCube;
	cube->width = 1.0f;
	cube->height = 1.0f;
	cube->depth = 1.0f;
	cube_separator->addChild(cube);

	root->addChild(cube_separator);
}
```

I am on debian linux and using the latest version of coin from github. I first created a discussion about this [here](https://github.com/orgs/coin3d/discussions/562) as I wasn't sure if this is a bug or wrong usage of the api from my side.

Dikluwe:
Fixes #563.

SoCube special-cases `SoMultiTextureEnabledElement::CUBEMAP` to send per-vertex 3D texture coordinates instead of 2D ones (needed since a cubemap sampler takes a direction vector, not a 2D UV pair) -- but it fed that path the same `sogl_cube_3dtexcoords[]` table used for `SoTexture3` (volumetric 3D textures), whose values are plain `[0,1]` per-axis corner coordinates with no relation to real cube geometry. A cubemap needs a real direction vector: the vertex's position relative to the cube's center, which for a cube centered at the origin is simply the vertex coordinate itself.

## Verification history (including a mistake I caught and fixed)

1. Rendered an `SoCube` with a solid, distinctly-colored `SoTextureCubeMap` (red/cyan/green/magenta/blue/yellow, one per face) via `SoOffscreenRenderer`, matching the issue's reproduction (a bare `SoCube` + `SoTextureCubeMap` with `filenames` set directly, no `SoTextureCoordinateReflectionMap` node). Before any fix, each face shows a smooth rainbow gradient uncorrelated with real cube geometry.
2. My first attempt used a second hand-written lookup table (the existing 0/1 values remapped to -1/+1), assumed to share `sogl_cube_3dtexcoords[]`'s per-vertex ordering. It didn't -- cross-checking against `sogl_generate_cube_vertices()`'s actual sign pattern showed the hand-written table's signs didn't match real vertex positions, so it was just as geometrically wrong as the original, in a different pattern. I built an independent, **Coin-free reference** (plain GLX + fixed-function OpenGL, same textures and camera, feeding `glTexCoord3fv()` the literal vertex position) and it did not match my first attempt's output.
3. Landed on using the vertex position directly (`varray[*iptr]`, already computed for `glVertex3fv()` right below it) instead of a second table -- this can't have a table/vertex-order mismatch by construction, and its output now matches the independent OpenGL reference exactly.

| Before (bug) | After (fix) | Plain OpenGL (independent reference) |
|---|---|---|
| Rainbow gradient uncorrelated with cube geometry | Solid color per face, blending at corners/edges | Same blending pattern as the fix |

Some gradient blending remains at the actual corners/edges in both the fix and the plain-OpenGL reference -- that's the real, unavoidable behavior of per-vertex (non-texgen) cubemap sampling on an unsubdivided 8-vertex cube with non-continuous face colors, not a residual bug; an ordinary photographic skybox image (continuous across its face boundaries by construction, unlike this synthetic test) won't show it.

---
**Verified:** Full clean rebuild under GCC and clang (Linux) with `-Wall -Wextra -Wpedantic`, 0 new warnings. `ctest`/`CoinTests` pass (327 tests) under a plain build and under a Debug build with AddressSanitizer/UndefinedBehaviorSanitizer, no new findings. The offscreen-render reproduction and independent plain-OpenGL reference described above confirm the visual fix directly.

**Not yet tested on Windows/MSVC** -- opening as a draft until that's verified.

### DIFF
--- a/src/rendering/SoGL.cpp
+++ b/src/rendering/SoGL.cpp
@@ -771,6 +771,29 @@ static float sogl_cube_3dtexcoords[][3] =
   {0.0f, 0.0f, 0.0f}
 };
 
+// SoTextureCubeMap needs a direction *vector* per vertex (so a flat
+// face samples one, constant, cubemap face -- the vector just needs
+// to point outward from the cube's center through that vertex), not
+// sogl_cube_3dtexcoords' [0,1] volume-texture coordinates above: with
+// those, adjacent corners of the same face differ in one component
+// (e.g. {1,1,1} next to {1,1,0}), so the GPU linearly interpolates
+// between two different cubemap faces' texels across what should be
+// a single uniform-colored face. Same per-vertex ordering as
+// sogl_cube_3dtexcoords, just remapped elementwise from [0,1] to
+// [-1,1] (0 -> -1, 1 -> +1) so each face's four corners agree on the
+// component that identifies it.
+static float sogl_cube_cubemaptexcoords[][3] =
+{
+  {1.0f, 1.0f, 1.0f},
+  {1.0f, 1.0f, -1.0f},
+  {1.0f, -1.0f, 1.0f},
+  {1.0f, -1.0f, -1.0f},
+  {-1.0f, 1.0f, 1.0f},
+  {-1.0f, 1.0f, -1.0f},
+  {-1.0f, -1.0f, 1.0f},
+  {-1.0f, -1.0f, -1.0f}
+};
+
 static float sogl_cube_normals[] =
 {
   0.0f, 0.0f, 1.0f,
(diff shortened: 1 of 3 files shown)


Dikluwe:
Fechando: fui aberto sem autorização prévia do autor. Será reaberto/recriado quando apropriado.

PR #703, opened 2026-09-07
Fix missing GL_UNPACK_ALIGNMENT in cubemap texture upload

Dikluwe:
Set GL_UNPACK_ALIGNMENT to 1 for cubemap face uploads so RGB rows whose byte count is not divisible by four are read with the correct stride. The rendering reproducer now releases local objects before Coin shutdown.

## Validation


veelo:
Thank you for your contribution!

PR #704, opened 2026-09-07
Fix SoTextureCubeMap distortion on SoCube (issue #563)

### LINKED ISSUE #563: SoTextureCubeMap incorrect UV mapping

Dikluwe:
Fixes #563.

## Summary

SoCube previously sent its volumetric 3D texture-coordinate table when cubemap texturing was active. Those coordinates are unit-cube positions, not directions from the cube center, and therefore distort cubemap lookup.

This change sends each cube vertex position as the cubemap direction while retaining the existing 3D texture behavior. It is limited to the SoCube rendering path and does not alter texture upload; the separate unpack-alignment defect is handled by #703.

## Regression evidence

The original validation rendered an SoCube with six distinctly colored cubemap faces through SoOffscreenRenderer and compared it with an independent fixed-function GLX reference using the same transform pipeline. The corrected image was pixel-identical to the reference.

## Current validation

The cubemap visual regression is not part of routine CoinTests and was not rerun in this update. The sanitizer run also observes a pre-existing SoReorganizeAction downcast diagnostic in the unrelated STL round-trip test when UBSan aborts on the first finding.

Windows, macOS and real-context rendering remain CI or manual validation gaps. The PR remains a draft until those results are reviewed.

### DIFF
--- a/src/rendering/SoGL.cpp
+++ b/src/rendering/SoGL.cpp
@@ -835,7 +835,16 @@ sogl_render_cube(const float width,
     if (flags & SOGL_MATERIAL_PER_PART)
       material->send(i, TRUE);
     for (int j = 0; j < 4; j++) {
-      if (flags & SOGL_NEED_3DTEXCOORDS) {
+      if (flags & SOGL_NEED_CUBEMAPTEXCOORDS) {
+        // The direction vector a cubemap lookup needs is just the
+        // vertex's own position relative to the cube's center (the
+        // GPU normalizes it, so magnitude doesn't matter) -- using
+        // the real vertex coordinates instead of a separately
+        // hand-maintained table rules out a table/vertex-order
+        // mismatch by construction.
+        glTexCoord3fv((const GLfloat*)&varray[*iptr]);
+      }
+      else if (flags & SOGL_NEED_3DTEXCOORDS) {
         glTexCoord3fv(sogl_cube_3dtexcoords[*iptr]);
       }
       else if (flags & SOGL_NEED_TEXCOORDS) {
--- a/src/rendering/SoGL.h
+++ b/src/rendering/SoGL.h
@@ -63,6 +63,7 @@ class SbVec2f;
 #define SOGL_NEED_TEXCOORDS      0x20
 #define SOGL_NEED_3DTEXCOORDS    0x40
 #define SOGL_NEED_MULTITEXCOORDS 0x80 // internal
+#define SOGL_NEED_CUBEMAPTEXCOORDS 0x100
 
 // Convenience function for access to OpenGL wrapper from an SoState
 // pointer.
--- a/src/shapenodes/SoCube.cpp
+++ b/src/shapenodes/SoCube.cpp
@@ -173,7 +173,7 @@ SoCube::GLRender(SoGLRenderAction * action)
       flags |= SOGL_NEED_TEXCOORDS;
       break;
     case SoMultiTextureEnabledElement::CUBEMAP:
-      flags |= SOGL_NEED_3DTEXCOORDS;
+      flags |= SOGL_NEED_CUBEMAPTEXCOORDS;
       break;
     }
   }


Dikluwe:
Re-ran the graphical validation using the exact six 2048x2048 JPEG cubemap faces and filename order from issue #563 (`right`, `left`, `top`, `bottom`, `front`, `back`), rather than synthetic solid-color or patterned faces.

The comparison isolates this PR's change: both libraries were built from the same PR head, with commit `936ae4ec7` reversed only for the **before** library. Both captures use the same input images, scene, camera, viewport and physical GPU.

Hardware renderer:

```
GL_VENDOR=AMD
GL_RENDERER=AMD Radeon Graphics (radeonsi, renoir, ACO, DRM 3.64, 6.17.0-42-generic)
GL_VERSION=4.6 (Compatibility Profile) Mesa 25.2.8-0ubuntu0.24.04.2
```

The comparison uses an orthographic isometric view, exposing three equally scaled cube faces. Before the fix, the cube's `[0,1]` 3D texture coordinates collapse the top face into stretched radial bands and select incorrect regions on the side faces. After the fix, vertex-position direction vectors restore the sky and the continuous photographic horizon. The two 640x640 captures differ in 196,236 of 409,600 pixels (normalized RMSE 0.173446).

The camera uses `-Y` as its up direction because `SoTextureCubeMap::filenames` documents the face order as `negX, posX, negY, posY, negZ, posZ`; in the reporter's filename sequence, `top.jpg` consequently maps to `-Y`. This keeps the source cubemap upright while preserving the exact face order used in the report.

The earlier synthetic tests remain useful as deterministic direction checks; this run confirms the same correction with the reporter's real photographic input.

veelo (approved):

veelo:
Thanks for fixing this, @Dikluwe! Please resolve conflicts so I can merge.

Dikluwe:
Thanks, @veelo. I checked the branch again against the current `master`. GitHub now reports the PR as clean and mergeable, so there are no conflicts left to resolve. The current head (`875fb0d7bd`) already contains the latest merge from `master` made on this branch, and all CI checks are green.

For completeness, the fix was also validated graphically in two independent ways: against a plain-OpenGL reference using the actual vertex-position direction vectors, and with the six 2048x2048 JPEG faces from issue #563. The latter comparison changed 196,236 of 409,600 pixels versus the broken path (normalized RMSE 0.173446) and restored the sky and continuous photographic horizon.

This should now be ready to merge.

veelo:
Well I don't know what the issue is then. GitHub says "This branch cannot be rebased due to conflicts". The "Rebase and merge" button is greyed out. There is a red button on the top right that says "Merge conflicts" but pressing it does not show the conflicts.

Perhaps GitHub is temporarily unwell, and we have to give it some time. Otherwise I will try to merge it offline.

PR #705, opened 2026-09-07
Fix multi-texture coordinate VBO registered under wrong texture unit

Dikluwe:
Register each texture-coordinate VBO for its actual texture unit instead of the wrong unit. The multi-texture reproducer now closes the full Coin lifecycle for leak checking.

## Validation


VolkerEnderlein:
Thanks for the contribution.

Issue #526, opened 2024-06-21
Multiple texture rendering problem with Coin 4


shanslow:
Hi folks,

We've encountered an issue with rendering multiple textures that was working fine previously on Coin 3.1.3. As shown below:

Rendered with Version 3.1.3
![image](https://github.com/coin3d/coin/assets/126081196/62d9c3c2-b17f-44c8-af3f-11698b909d59)

However, with version 4, it doesn’t get rendered as expected, please see snapshot 
below:
![image](https://github.com/coin3d/coin/assets/126081196/7cb82e64-cafe-4eb0-97f3-f22fba95e949)

Another interesting observation is, if we switch on the “Light Model node” in the scene, 
the scene will get rendered better apparently.

Scene file diff - with and without Light Model switched on
![image](https://github.com/coin3d/coin/assets/126081196/bc7a075d-5eba-4816-9112-51c6af832182)

Rendered with Version 4.0.0 with Light Model switched on
![image](https://github.com/coin3d/coin/assets/126081196/c348b350-f466-4c11-8477-9464725800a1)

I've included a small application example that highlights the problem when loading in the scene.iv from within the doc folder.  
[coinv4Issue.zip](https://github.com/user-attachments/files/15925412/coinv4Issue.zip)

In our full application we also see problems with the Z depth when rotating the model whether lighting is on or off. As we rotate, some of the distant faces pop in front of the nearer faces and vice versa.

Would you be able to take a look into this and see if it's a problem that requires a fix or if there's changes/settings we can make on our side to achieve the same behaviour we were seeing previously with Coin 3.1.3?

Thanks

Steve

VolkerEnderlein:
Unfortunately, I am not able to look into this over the next two weeks. Maybe @WizzerWorks or @veelo  can you have a look onto this isuue?

shanslow:
I don't really have the expertise to look into this, but have started to try and compare side by side runs in the debugger between 3.1.3 and 4.0.3. There's been a significant change in how the first unit texture is handled with everything now being managed my MultiTexture classes in v4 rather than plain Texture classes for unit zero and _then_ MultiTexture classes in v3. My suspicion is something to do with that but I don't have the expertise to know if settings are correct or not. I'm hoping to be able to find an indicator to where the key difference may lie and then offer up here for discussion.

shanslow:
[...]

PR #706, opened 2026-09-07
Fix off-by-one loop bound in SoGLMultiTextureImageElement::hasTransparency

Dikluwe:
## Summary

The static `hasTransparency(SoState *)` overload in
`SoGLMultiTextureImageElement` loops with `i <= unitdata.getLength()`,
one past the valid unit index range.

This is a cosmetic/robustness cleanup, not a behavior fix: the per-unit
`hasTransparency(int)` it calls already guards with
`unit < unitdata.getLength()` before touching the underlying array, so the
extra loop iteration was harmless in practice (it falls through the guard
and returns `FALSE`). Found incidentally while investigating #526; not the
cause of that issue.

Tightening the bound removes a redundant iteration and an off-by-one that
reads as riskier than it is on inspection — worth fixing even though it's
not user-visible.

## Test plan

- [x] Verified the loop bound is harmless: built a synthetic scene with 8
      texture units, exactly filling the internal `SbList`'s backing array
      to zero spare capacity, and ran under AddressSanitizer against both
      the unfixed and fixed code — no out-of-bounds access reported either
      way, confirming the inner bounds check is what actually protects
      this path.
- [x] Clean build with no new warnings on GCC 13 and Clang 18.
- [x] `ctest` passes on both compilers.

### DIFF
--- a/src/elements/GL/SoGLMultiTextureImageElement.cpp
+++ b/src/elements/GL/SoGLMultiTextureImageElement.cpp
@@ -294,7 +294,7 @@ SoGLMultiTextureImageElement::hasTransparency(SoState * state)
   const SoGLMultiTextureImageElement * elem = (const SoGLMultiTextureImageElement*)
     getConstElement(state, classStackIndex);
   
-  for (int i = 0; i <= PRIVATE(elem)->unitdata.getLength(); i++) {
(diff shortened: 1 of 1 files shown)


VolkerEnderlein:
Thanks for the fix. Your contribution is highly appreciated.

Issue #73, opened 2014-09-28
Memory leak during SoDB cleanup


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/73) by Konstantinos Dalamagkidis (Bitbucket: [dalai4bb](https://bitbucket.org/dalai4bb), ).**

Attachments: [patch-coin.diff](https://github.com/coin3d/attachments/blob/master/Coin3D_coin/issue73/patch-coin.diff)

----------------------------------------

Doing SoDB::init() and then SoDB::finish() should cleanup all memory according to the documentation of the latter, but in my case valgrind reported a memory leak.

The problem lies with the instantiation of the realTime SoGlobalField that is not properly deleted. I am not sure why the d'tor needs to be protected, but as it stands it is never called. I have attached a patch that makes the d'tor public and deletes the object in the callback.

I haven't tested this extensively so I don't know if it has any repercussions.

PR #709, opened 2026-09-07
Fix memory leak (and UAF on self-rename) in SoDB::renameGlobalField()

Dikluwe:
## Summary

`SoDB::renameGlobalField()` is documented as the way to discard a global field created via `SoDB::createGlobalField()`: "If \a to is an empty name, the \a from field gets deleted. If another global field already goes by the name \a to, that field will get deleted before the rename operation." Both of those "delete" branches only called `SoGlobalField::removeGlobalFieldContainer()`, which removes the item from the internal tracking list without ever touching its refcount (`SoGlobalField::initClass()` sets that list up with `addReferences(FALSE)` precisely so it doesn't) — so the field container was just dropped from the only list tracking it, never freed, regardless of its refcount.

This is the exact same bug class fixed for `SoDBP::removeRealTimeFieldCB()` in 2014 (Coin issue #73 / commit 3695870b88), whose own explanatory comment already correctly identified the mechanism. `renameGlobalField()` — the very function that fix's comment, and `SoGlobalField::clean()`'s shutdown warning, both point applications at as *the* correct way to discard a global field — was never itself touched, and has zero callers anywhere in Coin's own source tree or test suite to have ever exercised it.

Fix: call `unref()` on the field container instead (matching the 2014 fix exactly) — `~SoGlobalField()` already removes itself from `allcontainers`, so the explicit `removeGlobalFieldContainer()` call is both unnecessary and unsafe to keep.

Also corrected `SoGlobalField::removeGlobalFieldContainer()`'s own doc comment, which claimed (incorrectly, given `addReferences(FALSE)`) that it decreases the refcount — likely the source of the original mistake.

## Review fix (second commit)

A review of the first commit caught two more issues before this was ready:

1. **Use-after-free renaming a field to its own current name.** `renameGlobalField(name, name)` looks up `old` by the `to` name — when `from == to`, that lookup finds the very same container being renamed. `unref()`'ing it as if it were a distinct pre-existing entry destroys the object; the following `gf->setName(to)` then dereferences freed memory. Confirmed with ASan (heap-use-after-free). Fixed by returning immediately when `from == to` (a documented-contract no-op).
2. `testsuite/reproducers/sodb-renameglobalfield-leak/run.sh` had the same include-path bug already fixed on the `fix/null-dereference` branch (commit 8f2e897966) — recomputed the source include path via a second `dirname "$0"` lookup after already `cd`-ing into the reproducer's own directory, which breaks for relative-path invocation from elsewhere (e.g. the repo root, the exact usage shown in the script's own header comment). Applied the identical fix.

## Verification

New reproducer at `testsuite/reproducers/sodb-renameglobalfield-leak/`, exercising all three scenarios (delete via empty name, delete via renaming onto an existing name, self-rename no-op) via the documented `ref()`-then-delete ownership protocol (the same one `SoDB::init()` itself uses for `"realTime"`).

- Confirmed each bug reproduces cleanly before its respective fix and passes clean after, under GCC Debug, clang Debug, and Debug ASan+UBSan+LeakSanitizer together.
- Full `ctest` 100% pass in all three configurations (leak detection off for that broader run — an unrelated pre-existing leak in `SoBase_TestSuite::checkWriteWithMultiref`, which never `unref()`s its test scene graph, is out of scope here).

### DIFF
--- a/src/fields/SoGlobalField.cpp
+++ b/src/fields/SoGlobalField.cpp
@@ -198,9 +198,14 @@ SoGlobalField::addGlobalFieldContainer(SoGlobalField * fieldcontainer)
 
 // Remove the given global field from the internal list.
 //
-// Note that this will decrease the reference count of the
-// SoGlobalField node, causing it to be destructed unless it has been
-// ref()'ed outside of this class.
+// Note that this does *not* affect the reference count of the
+// SoGlobalField node: allcontainers is set up in initClass() with
+// addReferences(FALSE) precisely so that removing an item from it here
+// does not implicitly ref()/unref() -- callers that actually want the
+// fieldcontainer deleted need to unref() it themselves (which will, in
+// turn, remove it from this list via ~SoGlobalField() -- calling this
+// method first is both unnecessary and unsafe, since the destructor's
+// own removal would then find nothing to remove).
 void
 SoGlobalField::removeGlobalFieldContainer(SoGlobalField * fieldcontainer)
 {
--- a/src/misc/SoDB.cpp
+++ b/src/misc/SoDB.cpp
@@ -955,6 +955,13 @@ SoDB::renameGlobalField(const SbName & from, const SbName & to)
   }
 #endif // COIN_DEBUG
 
+  // Renaming a field to its own current name is a no-op -- and must be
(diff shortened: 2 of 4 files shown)


VolkerEnderlein:
Thanks for the contribution.

Issue #149, opened 2018-04-26
SbSphereSheetProjector.cpp has special characters which may cause bug


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/149) by zaxs0130 (Bitbucket: [zaxs0130](https://bitbucket.org/zaxs0130), GitHub: [zaxs0130](https://github.com/zaxs0130)).**

----------------------------------------

at line 192 and 200
![special characters](https://bitbucket.org/repo/8b7xA/images/198810812-Snipaste_2018-04-27_00-09-38.png)

Under some encoding, these special characters may lead to wrong EOLs like the image below and cause runtime error (bug).
![error](https://bitbucket.org/repo/8b7xA/images/3070711194-20180427001551.png)
The special character make line 192 and line 193 into one line (192), and cause the error below.

PR #710, opened 2026-09-07
Remove remaining non-ASCII characters from source comments

Dikluwe:
## Summary

Coin issue #149 (2018) reported that a non-ASCII character (a Latin-1 degree sign, U+00B0) sitting as the last character on a line in `SbSphereSheetProjector.cpp` made two lines merge into one under MSVC 2017 with a Chinese-Simplified system locale, producing a compile error. This is a known MSVC behavior: without a BOM or an explicit `/utf-8` (or `/source-charset:utf-8`) flag, source files are decoded using the active code page for non-Unicode programs, so a non-ASCII byte sequence near a line ending can, under a code page it wasn't written for, get reinterpreted such that the CR/LF bytes are consumed as part of a bogus multi-byte character — silently deleting the line break as far as the tokenizer is concerned.

The maintainer fixed only that one file (commit 3edeb8a332), reasoning "I changed only the file in question, as it is the only one with non ASCII characters at the end of line" — true for the reported crash, but he also wrote in the same issue thread: "I've seen some other files in Coin containing also non ASCII characters (especially when expressing formulae). Maybe it's time to get rid of it." That broader cleanup never happened.

23 files (106 lines) still had non-ASCII bytes — all valid UTF-8 today (no mixed-encoding issue currently), but the same MSVC code-page misinterpretation risk applies to any of them, not just ones sitting exactly at end-of-line: mostly hand-written formula comments using × (multiplication/cross product), · (dot product), ² (squared), ° (degree) and ± (plus-minus), plus a handful of one-off cases (two accented personal names, a font-bug comment quoting three Scandinavian letters, and three exotic letters used purely as fanciful placeholder symbols for x/y/z sub-expressions in a formula comment).

Replaced all of them with ASCII equivalents, matching conventions already used elsewhere in the very same comment blocks (e.g. `*` and `^2` both already appear alongside `×` and `²` in the same formula derivations) or the exact wording the 2019 fix used for the degree sign ("45 degree").

Auto-generated bitmap font tables under `src/fonts/*.icc` (glyph-index comments spanning the full Latin-1 range) were deliberately left out of scope: they are machine-generated, not hand-written prose, and their non-ASCII bytes are never the last character on a physical line (always followed by a closing `*/`), so the specific EOL-merge mechanism from #149 does not apply to them.

## Review fixes (second commit)

Two editorial issues found on review before this was ready:

1. 12 lines carried trailing whitespace preserved from the original text (`git diff --check` flagged them) — stripped.
2. Two formulas in `Extrusion.cpp` used `*` for what the surrounding prose explicitly calls a cross-product, inherited from the blanket `×` → `*` substitution (which was correct for the scalar-multiplication uses elsewhere but not this one). Changed to `cross(a, b)` notation to keep the operation unambiguous.

## Verification

Comment-only change (confirmed with a diff review — no non-comment lines changed). Byte-level scan confirms zero non-ASCII bytes remain in the touched files. Clean build + `ctest` pass under both GCC and clang Debug.

### DIFF
--- a/src/base/SbCylinder.cpp
+++ b/src/base/SbCylinder.cpp
@@ -180,7 +180,7 @@ SbCylinder::intersect(const SbLine& l, SbVec3f& enter, SbVec3f& exit) const
   // be written as
   //                     Qc = Pc + x*U0 + y*U1 + z*U2
   //
-  // The points on the cylinder is then those matching x²+y²=r².
+  // The points on the cylinder is then those matching x^2+y^2=r^2.
   //
   // For the ray, we project it into the coordinate system by
   // rewriting the line equation   Qr = Pr + t*Dr  as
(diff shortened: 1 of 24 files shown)


VolkerEnderlein:
Thanks for the contribution.

Issue #173, opened 2019-03-09
coin fails to build vs spidermonkey-52.9.1_pre1


VolkerEnderlein:
**[Original report](https://bitbucket.org/Coin3D/coin/issue/173) by Fabio Rossi (Bitbucket: [efferre79](https://bitbucket.org/efferre79), GitHub: [efferre79](https://github.com/efferre79)).**

Attachments: [coin-9999:20190309-121941.log.gz](https://github.com/coin3d/attachments/blob/master/Coin3D_coin/issue173/coin-9999%3A20190309-121941.log.gz)

----------------------------------------

The first errors start as

```
#!c++

[ 26%] Building CXX object src/base/CMakeFiles/base.dir/SbRotation.cpp.o
cd /var/tmp/portage/media-libs/coin-9999/work/coin-9999_build/src/base && /usr/bin/x86_64-pc-linux-gnu-g
In file included from m/usr/include/mozjs-52/jsalloc.h:17:0,
                 from m/usr/include/mozjs-52/jsapi.h:26,
                 from m/var/tmp/portage/media-libs/coin-9999/work/Coin-9999/src/glue/spidermonkey.cpp:58
m/usr/include/mozjs-52/js/TypeDecls.h:25:8: error: using typedef-name ‘mJSContext’ after ‘mstruct’
 struct JSContext;
        ^~~~~~~~~
In file included from m/var/tmp/portage/media-libs/coin-9999/work/Coin-9999/src/glue/spidermonkey.cpp:34
m/var/tmp/portage/media-libs/coin-9999/work/Coin-9999/include/Inventor/C/glue/spidermonkey.h:104:14: not
 typedef void JSContext;
              ^~~~~~~~~
In file included from m/usr/include/mozjs-52/jsalloc.h:17:0,
                 from m/usr/include/mozjs-52/jsapi.h:26,
                 from m/var/tmp/portage/media-libs/coin-9999/work/Coin-9999/src/glue/spidermonkey.cpp:58
m/usr/include/mozjs-52/js/TypeDecls.h:26:7: error: using typedef-name ‘mJSFunction’ after ‘mclass’
 class JSFunction;
       ^~~~~~~~~~
```

I am attaching the full build log

Dikluwe:
[...]

PR #713, opened 2026-09-08
Fail cleanly at configure time for SpiderMonkey link-time linking (issue #173)

Dikluwe:
## Summary

Coin issue #173 reported that building with `SPIDERMONKEY_RUNTIME_LINKING=OFF` (link-time linking against a real SpiderMonkey, as opposed to the default run-time `dlopen` path) fails to compile against any modern SpiderMonkey/mozjs with hundreds of conflicting-declaration errors (`JSContext`, `JSType`, `jsid`, `JSGCStatus`, ... all redefined).

Root-caused by actually reproducing it: downloaded and extracted the real `libmozjs-102-dev` package (no root needed, via `apt-get download` + `dpkg-deb -x`) and compiled `<jsapi.h>` followed by `<Inventor/C/glue/spidermonkey.h>` directly. Confirmed the reported errors verbatim. Coin's own header guards its fallback declarations with `#ifndef JSVERSION_IS_ECMA`, meant to detect "the real jsapi.h was already included" -- but that macro was removed from SpiderMonkey upstream in 2013 and no longer exists in any release available today, so the guard always takes the fallback branch and always redeclares types the real header already provides.

I tried the "obvious" fix first (gate the fallback block on the existing, CMake-provided `HAVE_SPIDERMONKEY_VIA_LINKTIME_LINKING` macro instead) and it does **not** work: doing so only clears the first layer of conflicts. A second, much larger wave of errors follows, because Coin's function-pointer typedefs (`JS_EvaluateScript_t`, `JS_SetProperty_t`, ...) are declared in terms of types (`JSBool`, `uintN`, `jsval`-as-integer, `intN`) that don't just live under a different guard in modern SpiderMonkey -- they don't exist at all anymore (confirmed by grepping the real mozjs-102 headers: zero matches for any of them). The entire glue in `src/glue/spidermonkey.cpp` targets the SpiderMonkey C API as it existed around 2007, which is exactly what #673 (merged into this repo) already independently found and documented while fixing the run-time (`dlopen`) path's availability check -- that PR explicitly scoped out porting this glue to a current SpiderMonkey embedding API as a separate, much larger decision.

Given that, the fix here is to make the doomed configuration fail loudly and immediately at `cmake` configure time -- before hundreds of compiler errors -- with a message explaining why and pointing at the working default, rather than attempting an actual port of the glue.

While testing the "SpiderMonkey not found" case for comparison, also found (and fixed) two separate, pre-existing bugs in `cmake/FindSpiderMonkey.cmake`, confirmed to reproduce identically on unmodified master:
- It hard-crashes (`file STRINGS given unknown argument`) instead of reporting `SPIDERMONKEY_FOUND=FALSE` when no SpiderMonkey is present at all and `SPIDERMONKEY_RUNTIME_LINKING=OFF`.
- If a SpiderMonkey's `jsversion.h`/`jsconfig.h` exists but doesn't define `JS_VERSION`, `string(REGEX REPLACE ...)` is called with a missing argument and CMake rejects it outright. Also, whenever the version couldn't be determined by any path, `SPIDERMONKEY_VERSION` was left holding the raw include-dir path, which then leaked into `find_library()`'s `NAMES` list instead of falling back to the plain, unversioned names already there.

## Test plan

- Reproduced the original #173 compile failure with a real `libmozjs-102-dev` package.
- Verified the new `FATAL_ERROR` fires correctly against that same package when `SPIDERMONKEY_RUNTIME_LINKING=OFF`, with no compiler ever invoked.
- Verified the default (`SPIDERMONKEY_RUNTIME_LINKING=ON`) and `COIN_HAVE_JAVASCRIPT=OFF` configurations are unaffected.
- Verified all `FindSpiderMonkey.cmake` fixes with synthetic include directories isolating each case: SpiderMonkey missing entirely, a version embedded in the include path, a version read from `jsversion.h`, `jsversion.h` present without `JS_VERSION`, and neither version file present -- each configures cleanly with the correct `SPIDERMONKEY_VERSION` (or a clean fallback), no crashes.
- Full clean rebuild under GCC and Clang (Debug), `ctest` (1/1 pass) on both, and a Debug build under AddressSanitizer/UndefinedBehaviorSanitizer running the full `CoinTests` suite (330 tests, 83677 checks, zero UBSan findings, the same 207 pre-existing leak allocations as the established baseline from #673) -- across all three commits on this branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,16 +491,25 @@ if(HAVE_VRML97 AND COIN_HAVE_JAVASCRIPT)
   if(NOT SPIDERMONKEY_RUNTIME_LINKING)
     find_package(SpiderMonkey)
     if(SPIDERMONKEY_FOUND)
-      set(HAVE_SPIDERMONKEY_VIA_LINKTIME_LINKING 1)
-      set(HAVE_SPIDERMONKEY_H 1)
-      if(NOT TARGET SpiderMonkey::SpiderMonkey)
-        add_library(SpiderMonkey::SpiderMonkey UNKNOWN IMPORTED)
-        set_target_properties(SpiderMonkey::SpiderMonkey PROPERTIES
-          IMPORTED_LOCATION "${SPIDERMONKEY_LIBRARIES}"
-          INTERFACE_INCLUDE_DIRECTORIES "${SPIDERMONKEY_INCLUDE_DIRS}"
-        )
-      endif()
-      list(APPEND COIN_TARGET_LINK_LIBRARIES SpiderMonkey::SpiderMonkey)
+      message(FATAL_ERROR
+        "SPIDERMONKEY_RUNTIME_LINKING=OFF (link-time SpiderMonkey linking) "
+        "found a SpiderMonkey at '${SPIDERMONKEY_INCLUDE_DIR}', but this "
+        "configuration cannot work: Coin's SpiderMonkey glue "
+        "(src/glue/spidermonkey.cpp, include/Inventor/C/glue/spidermonkey.h) "
+        "targets the SpiderMonkey C API as it existed around 2007 (JSBool, "
+        "jsval as a plain integer, no JS::Rooted<>/JS::PropertyKey), and no "
+        "SpiderMonkey release still available today exports that API -- "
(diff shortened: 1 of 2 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #714, opened 2026-09-09
Fix SoPath/SoFullPath downcast undefined behavior across the codebase

Dikluwe:
## Summary

Replace undefined `SoPath`-to-`SoFullPath` downcasts with a composition-based full-path view, while preserving the existing Coin 4 class hierarchy and ABI.

### Public migration API

Code that needs to inspect the complete stored path, including hidden nodekit children, should use `SoPath::fullPath()`:

```cpp
// Historical idiom: undefined unless path points to an object that was
// actually constructed as SoFullPath or one of its subclasses.
SoNode * tail = static_cast<SoFullPath *>(path)->getTail();

// Preferred Coin API: valid for every SoPath.
SoNode * tail = path->fullPath().getTail();
```

The complete migration is:

| Historical call | Preferred call |
| --- | --- |
| `static_cast<SoFullPath *>(path)->getLength()` | `path->fullPath().getLength()` |
| `static_cast<SoFullPath *>(path)->getTail()` | `path->fullPath().getTail()` |
| `static_cast<SoFullPath *>(path)->getNodeFromTail(i)` | `path->fullPath().getNodeFromTail(i)` |
| `static_cast<SoFullPath *>(path)->getIndexFromTail(i)` | `path->fullPath().getIndexFromTail(i)` |

`SoFullPathView` is a borrowed, one-pointer view over `const SoPath`. Creating it performs no allocation and does not change reference counts. It exposes only the complete-path accessors and remains valid for as long as the viewed `SoPath` remains valid.

The visible-path API on `SoPath` is unchanged: `getLength()`, `getTail()`, `getNodeFromTail()`, and `getIndexFromTail()` still stop at hidden children as before.

## Why the historical cast is undefined

`SoPath` already owns the complete node and index arrays. `SoFullPath` adds no instance storage; it hides several non-virtual `SoPath` methods with equivalents that inspect the complete arrays. Nevertheless, identical layout does not make a base-class object an instance of a derived class.

A base-to-derived pointer conversion is valid only when the base pointer designates the base subobject of an object of the derived type. Treating an arbitrary `SoPath` as an `SoFullPath` therefore has undefined behavior:

```cpp
SoTempPath * temp = ...;
SoFullPath * valid = temp; // Valid: genuine inheritance.

SoPath * path = new SoPath(...);
SoFullPath * invalid = static_cast<SoFullPath *>(path); // Undefined behavior.
```

This object-identity requirement was already present in the November 1997 C++ working paper and remains in the current standard:

- Historical wording: https://open-std.org/jtc1/sc22/WG21/docs/wp/html/nov97-2/expr.html
- Current wording: https://eel.is/c++draft/expr.static.cast

`dynamic_cast` would safely reject the second case, but it would not provide access to the complete-path interface. `reinterpret_cast` and C-style casts do not create a derived object either.

## Design and compatibility

- `SoPath::fullPath()` returns the new composition-based `SoFullPathView`.
- `SoFullPath` remains available for source and binary compatibility.
- `SoTempPath` continues to derive genuinely from `SoFullPath`.
- Legacy `SoFullPath` accessors delegate to `SoFullPathView`, making the view the single implementation of complete-path access.
- No transitional public `SoPath::getFull*()` API is exposed, so downstream users need only one migration.
- `SoPath` layout and vtable are unchanged.
- No existing public function or variable is removed or changed.

The existing class is not deprecated in Coin 4 because it remains a real base of `SoTempPath`. What is unsupported is using an arbitrary `SoPath` object as though it had been constructed as `SoFullPath`.

Issue #741 remains focused on the Coin 5 design choice: retain `SoFullPathView` as the modern name, or redefine `SoFullPath` as the composition-based view and move `SoTempPath` directly under `SoPath`.

## Historical and downstream context

The cast pattern is inherited from the Open Inventor API rather than being accidental downstream misuse. Coin documentation historically described `SoFullPath` as an extended interface over layout-compatible `SoPath` storage and recommended casting to it. Coin also used the pattern internally.

The managed Open Inventor API provides a modern precedent for composition by exposing a `FullPath` property and calls such as `path.FullPath.GetTail()`:

- Current Open Inventor C++ definition: https://developer.openinventor.com/refmans/2026.1/RefManCpp/_so_path_8h_source.html
- Managed `FullPath` property: https://developer.openinventor.com/refmans/latest/RefManNet/html/e4f6fe7d-de3a-309f-550d-76b7a16ba764.htm

The problem is observable in FreeCAD. FreeCAD PR [#28746](https://github.com/FreeCAD/FreeCAD/pull/28746), merged in April 2026, added [`SoFullPathHelper.h`](https://github.com/FreeCAD/FreeCAD/blob/main/src/Gui/SoFullPathHelper.h) to suppress the sanitizer diagnostic around this cast. After adopting a Coin release containing this PR, FreeCAD can migrate those calls to `path->fullPath()` and remove that sanitizer-suppressed helper.

## Scope

- Migrates Coin's internal unsafe `SoPath`-to-`SoFullPath` casts to `fullPath()`.
- Preserves legitimate `SoFullPath`/`SoTempPath` inheritance.
- Updates public documentation and examples that recommended the historical cast.
- Adds coverage for hidden children, plain `SoPath` objects, genuine `SoFullPath` subclasses, empty paths, invalid indices, profiling lookup, and the existing STL reorganization path.
- Adds an unreleased `NEWS` entry documenting the preferred migration API.

The branch deliberately combines the corrected inclusive traversal bound from #716 with the full-path-view migration in `SbProfilingData`, and preserves the newer `coin_safe_cast` parent validation in `SoReorganizeAction`.

## Validation

- `CoinTests`: 344 tests and 83,816 checks passed.
- CTest: 8/8 passed.
- GCC and Clang builds passed.
- AddressSanitizer and UndefinedBehaviorSanitizer validation passed for the migrated paths.
- `git diff --check`: clean.
- `abidiff` against Coin 4.0.10: 0 removed and 0 changed functions or variables; additions only.
- GitHub CI passed on Ubuntu, Windows, and macOS with the legacy renderer both enabled and disabled.
- CodeQL and the ABI compatibility workflow passed.

This is the Coin 4 compatibility repair. It removes the undefined downcasts without requiring an ABI break and gives downstream users the same migration API intended to survive into Coin 5.

### DIFF
--- a/include/Inventor/SoFullPath.h
+++ b/include/Inventor/SoFullPath.h
@@ -56,7 +56,7 @@ public:
 inline void
 SoFullPath::pop(void)
 {
-  this->truncate(this->nodes.getLength() - 1);
+  this->truncate(this->getFullLength() - 1);
 }
 
 #endif // !COIN_SOFULLPATH_H
--- a/include/Inventor/SoPath.h
+++ b/include/Inventor/SoPath.h
@@ -82,12 +82,7 @@ public:
   // "Full" variants of getTail()/getNodeFromTail()/getIndexFromTail()/
   // getLength(): unlike the methods above, these count hidden children
   // (e.g. nodekit-internal nodes) instead of stopping at the first one.
-  // This is the same data SoFullPath's identically-named methods expose
-  // via a reinterpret_cast<SoFullPath*>(path) of a plain SoPath -- these
-  // provide the same thing directly on SoPath, without the cast (which
-  // is undefined behavior whenever the SoPath instance was never
-  // actually constructed as a SoFullPath, as it usually isn't).
-  // SoFullPath itself now delegates to these.
+  // Use these to inspect a complete SoPath without casting to SoFullPath.
   SoNode * getFullTail(void) const;
   SoNode * getFullNodeFromTail(const int index) const;
   int getFullIndexFromTail(const int index) const;
--- a/include/Inventor/SoPath.h
+++ b/include/Inventor/SoPath.h
@@ -79,6 +79,20 @@ public:
   int getLength(void) const;
   void truncate(const int length);
 
+  // "Full" variants of getTail()/getNodeFromTail()/getIndexFromTail()/
+  // getLength(): unlike the methods above, these count hidden children
+  // (e.g. nodekit-internal nodes) instead of stopping at the first one.
+  // This is the same data SoFullPath's identically-named methods expose
+  // via a reinterpret_cast<SoFullPath*>(path) of a plain SoPath -- these
+  // provide the same thing directly on SoPath, without the cast (which
+  // is undefined behavior whenever the SoPath instance was never
+  // actually constructed as a SoFullPath, as it usually isn't).
+  // SoFullPath itself now delegates to these.
+  SoNode * getFullTail(void) const;
+  SoNode * getFullNodeFromTail(const int index) const;
+  int getFullIndexFromTail(const int index) const;
+  int getFullLength(void) const;
+
   int findFork(const SoPath * const path) const;
   int findNode(const SoNode * const node) const;
 
@@ -107,7 +121,6 @@ private:
   static void cleanupClass(void);
   static void * createInstance(void);
   void append(SoNode * const node, const int index);
-  int getFullLength(void) const;
   void truncate(const int length, const SbBool donotify);
   SbBool readInstance(SoInput * in, unsigned short flags) override;
   void setFirstHidden(void);
@@ -119,7 +132,6 @@ private:
   SbBool firsthiddendirty;
   static SoType classTypeId;
 
-  friend class SoFullPath;
   friend class SoNodeKitPath;
   friend class SoAction;
   friend class SoTempPath;
--- a/include/Inventor/annex/Profiler/SbProfilingData.h
+++ b/include/Inventor/annex/Profiler/SbProfilingData.h
@@ -42,7 +42,6 @@
 
 class SoNode;
 class SoPath;
-class SoFullPath;
 class SbProfilingDataP;
 
 typedef void * SbProfilingNodeKey; // void since it should not be dereferenced
@@ -144,13 +143,13 @@ private:
 
   void constructorInit(void);
 
-  SbBool isPathMatch(const SoFullPath * path, int pathlen, int idx);
+  SbBool isPathMatch(const SoPath * path, int pathlen, int idx);
 
-  int getIndexCreate(const SoFullPath * path, int pathlen);
+  int getIndexCreate(const SoPath * path, int pathlen);
   int getIndexNoCreate(const SoPath * path, int pathlen) const;
-  int getIndexForwardCreate(const SoFullPath * path, int pathlen,
+  int getIndexForwardCreate(const SoPath * path, int pathlen,
                             int parentindex);
-  int getIndexForwardNoCreate(const SoFullPath * path, int pathlen,
+  int getIndexForwardNoCreate(const SoPath * path, int pathlen,
                               int parentindex) const;
 
 }; // SbProfilingData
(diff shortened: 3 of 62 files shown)


VolkerEnderlein:
@Dikluwe I am out in the middle of nowhere with very limited internet access for the next three weeks and therefore cannot react properly on the PRs. @veelo or @WizzerWorks can you have a look onto the open PRs? Thanks in advance.

veelo:
@Lgt2x since you have expressed your concern regarding FreeCAD before, perhaps you can have a look at this one. It directly addresses a recent address sanitizer suppressor in FreeCAD, see above.

I understand that downcasting SoPath objects to SoFullPath representations is undefined behaviour in the adopted standard, and that it is a wart in the public API. The original intention of this design was probably to put the most frequently desired behaviour up front and hide the special cases behind an extra step.

This PR folds SoFullPath function equivalents into SoPath while retaining backwards functionality. This breaks that design intent, but I don't think it is that important. It is different from the [.NET property](https://developer.openinventor.com/refmans/latest/RefManNet/html/166096df-a38b-0cde-3c71-f0cb7ffa8782.htm), but something like that is proposed for Coin 5 in #741.

I am prepared to merge this, but will allow some time for input from the FreeCAD folks. @WizzerWorks?

veelo:
I have slept on this, and I don't think we can add `getFull*()` here and remove it in Coin 5 in favour of a view based solution. Users that want to get rid of the UB in Coin 4 would have to change their code twice.

So I think there are two options:
1) We accept the unified class as the final solution, syntax `path.getFullTail()` etc. `SoFullPath` can remain for backwards compatibility but gets deprecated. With this, Coin would deviate from other Open Inventor implementations, as far as I know.
2) We go directly for the view based solution, syntax `path.fullPath().getTail()`. `SoFullPath` is deprecated, `getFull*()` is implemented but deprecated. In Coin 4 `fullPath` is implemented on top of `getFull*()`, in Coin 5 we can choose to remove `SoFullPath` and `getFull*()` and reimplement `fullPath` proper. This aligns with the .NET Open Inventor implementation.

Question is: is limiting access to hidden children so important that it warrants a view? 1. is cleaner, 2. retains more of the original design intent.

Dikluwe:
I implemented a refined version of option 2 in commit `ce85d849c7`.

The public migration API is now:

```cpp
path->fullPath().getTail();
```

The main difference from the transitional form described above is that Coin 4 now gets a real composition-based `SoFullPathView` immediately. It is a borrowed, one-pointer view over `const SoPath`, with no allocation or reference-count change. We do not expose public deprecated `getFull*()` methods for `fullPath()` to call through, so downstream users have only one migration and Coin 5 does not need to remove an intermediate public API.

The legacy direction is reversed: `SoFullPath` remains ABI compatible and its accessors delegate to `SoFullPathView`. I did not mark the class itself deprecated in Coin 4 because `SoTempPath` genuinely derives from it; only using an arbitrary `SoPath` as an `SoFullPath` is unsupported. The historical internal `SoPath::getFullLength()` is private again.

All migrated internal call sites now use `fullPath()`. This retains the original visible/full interface separation while removing the invalid base-to-derived casts.

Validation:

- `CoinTests`: 344 tests and 83,816 checks passed.
- CTest: 8/8 passed.
- `git diff --check`: clean.
- `abidiff` against Coin 4.0.10: 0 removed and 0 changed functions or variables; additions only. `SoPath` layout and vtable are unchanged.

This leaves #741 focused on the actual Coin 5 choice: keep `SoFullPathView` as the modern name, or redefine `SoFullPath` as the composition-based view and move `SoTempPath` directly under `SoPath`.

veelo:
Thank you very much for quick adoption, I think we are in a good place with this. We should probably document and communicate the new public migration API in the next release notes, and I am not quite sure where to leave this information in the mean time. Could you update the PR summary to this API to help with this? Thanks!

veelo:
Please also add a few lines at the top of the NEWS file in the project root explaining to users what the preferred API is that removes UB. When we prepare the next release we will find those lines and incorporate them into the release summary. Feel free te leave a reference to this PR.

Dikluwe:
Done in commit `6a47c2df7e`. I updated the PR summary to present `path->fullPath()` as the public migration API and added an unreleased entry at the top of `NEWS` explaining how to replace the undefined `SoPath`-to-`SoFullPath` downcast. Thanks!

veelo:
Excellent work. Thank you very much!

PR #715, opened 2026-09-09
Fix uninitialized fbc_cnt read in glxglue_context_create_pbuffer()

Dikluwe:
Initialize fbc_cnt to zero before glXChooseFBConfig. If the chooser fails without writing its count, the existing fallback now reads a defined value.

Keep the production change minimal. The reproducer includes the actual implementation translation unit with four mocked chooser outcomes, rather than exporting a test-only helper from Coin. Document compiler instrumentation and the distinction between mocks and a real GLX context.

## Validation


VolkerEnderlein:
Thanks for the contribution.

PR #718, opened 2026-09-09
Fix SoCallbackList function-pointer-type undefined behavior

Dikluwe:
Typed callbacks in SoDragger, SoGLRenderAction and SoSelection were invoked through an incompatible generic function-pointer signature. Store and dispatch the private lists with their real types; preserve SoSelection's protected SoCallbackList members using owned, type-correct adapters. Identity casts are used only for registration matching.

Add three CoinTests cases covering copies, assignment, owner destruction, mixed registrations, last-duplicate removal and mutation during dispatch. The compatibility client uses all public headers from fixed pre-change revision c27cf9a72d302c8c1ee3367a8865a603ea26622a and keeps checks active with NDEBUG. Reproducers now clean up Coin after local objects.

## Validation


VolkerEnderlein:
Thanks for the contribution.

Issue #534, opened 2024-11-12
Incorrect rendering if graph contains many nested SoTransform nodes


wwmayer:
Hi Coin team,

if a scene graph contains many nested SoTransform nodes then it can happen that only a small part is actually rendered. 

Interesting in this context is that this happens when using IndexedFacetSet nodes for the geometry but when using SoCube instead everything is correct. Also when converting the IndexedFacetSet into VRML the rendering is correct, too.

Here is the code to create a simple viewer:
```
#include <GL/gl.h>
#include <Inventor/SoDB.h>
#include <Inventor/SoInteraction.h>
#include <Inventor/SoInput.h>
#include <Inventor/nodes/SoSeparator.h>

#include <Inventor/Qt/SoQt.h>
#include <Inventor/Qt/viewers/SoQtExaminerViewer.h>


int main(int argc, char** argv)
{
    // built with:
    // g++ view.cpp -fPIC -I/usr/include -I/usr/include/x86_64-linux-gnu/qt5  
    // -I/usr/include/x86_64-linux-gnu/qt5/QtCore
    // -L/usr/lib/x86_64-linux-gnu -lCoin -lSoQt -lGL -lQt5Core -o viewer
    //

    if (argc < 2) {
        return -1;
    }

    SoDB::init();
    SoInteraction::init();

    SoSeparator* root = new SoSeparator();
    root->ref();

    SoInput myInput;
    myInput.openFile(argv[1]);
    SoNode* geomObject = SoDB::readAll(&myInput);
    if (geomObject) {
        root->addChild(geomObject);
    }

    QWidget* mainwin = SoQt::init(argc, argv, argv[0]);
    SoQtExaminerViewer* eviewer = new SoQtExaminerViewer(mainwin);
    eviewer->setSceneGraph(root);

    eviewer->show();

    SoQt::show(mainwin);
    SoQt::mainLoop();

    delete eviewer;
    root->unref();
    SoQt::done();
    return 0;
}
```
[mesh.zip](https://github.com/user-attachments/files/17716893/mesh.zip)

The attached zip file contains the files: cube.iv, mesh.iv and mesh.wrl. All of them have ~100 nested transform nodes.

Start the viewer with:
* `./viewer cube.iv` and the rendering is correct
* `./viewer mesh.iv` and the rendering is incorrect. It only shows ~30 of the 100 meshes. When rotating the scene you may observe some flickering where some further meshes are tmp. rendered
[...]

PR #720, opened 2026-09-10
Skip identity-transform matrix element updates in SoScale/SoRotation/SoRotationXYZ

Dikluwe:
Skip exact-identity matrix-element updates in `SoScale`, `SoRotation`, and `SoRotationXYZ`, following the existing `SoTransform` optimization. Non-identity values retain the existing behavior. This is an optimization, not a demonstrated fix for issue #534's visual corruption.

Add a portable `CoinTests` scenario shared with the standalone reproducer. For each node, it checks identity → non-identity → identity with independently specified matrices and bounding boxes. Caching is enabled once; a traversal counter verifies a rebuild after each change and reuse on an unchanged repeat. A 2×4×6 box makes stale rotation results observable. The standalone driver releases Coin globals after destroying its scene and actions.

## Validation

- Clean Linux Debug builds with GCC and Clang: CTest passed; 331 tests / 83,722 checks, including 45 assertions in the new scenario. Standalone reproducer passed with both compilers.
- Negative controls: reassigning the cache option before each measurement fails the reuse assertions; substituting the original box for subsequent measurements fails the expected-extents assertions for all three node types.
- Clang Debug ASan/UBSan instrumented both Coin and CoinTests, and separately the standalone driver: `-O0 -g1 -fsanitize=address,undefined -fno-omit-frame-pointer`, with `ASAN_OPTIONS=detect_leaks=1`, no suppressions. The standalone driver passed with leak detection enabled and UBSan set to halt on error. All CoinTests assertions passed with no UBSan reports, but CTest failed under LeakSanitizer: 11,552 bytes in 207 allocations, exactly matching a comparison run of the original 330-test suite from the published PR. The full sanitizer suite is not reported as a clean pass.
- Reviewed the final diff, documentation, and relevant conditional code; `git diff --check` passed. The new test has no platform or optional-feature guards and is extracted from the `COIN_TEST_SUITE` block into CoinTests.

## Limitations

The new scenario exercises matrix and bounding-box actions, not GL rendering or render-cache behavior. Windows/macOS and optional configurations were not executed locally. The existing full-suite leaks remain outside these test corrections. No benchmark or reproduction of issue #534's visual symptom is claimed.

### DIFF
--- a/src/nodes/SoRotation.cpp
+++ b/src/nodes/SoRotation.cpp
@@ -134,8 +134,17 @@ void
 SoRotation::doAction(SoAction * action)
 {
   if (!this->rotation.isIgnored()) {
-    SoModelMatrixElement::rotateBy(action->getState(), this,
-                                   this->rotation.getValue());
+    const SbRotation & rot = this->rotation.getValue();
+    // Skip the matrix element update entirely for an identity rotation
+    // -- same "no-op transform contributes no GL/matrix work"
+    // optimization SoTransform::doAction() got in issue #534, applied
+    // to this sibling node. isIgnored() is a separate, orthogonal
+    // check (an SoField connection/override flag) and is preserved
+    // above exactly as before; this adds a second, independent check
+    // on the field's *value*. Exact comparison on purpose.
+    if (rot != SbRotation::identity()) {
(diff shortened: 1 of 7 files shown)


VolkerEnderlein:
Thanks for the contribution.

PR #721, opened 2026-09-10
Fix SoTransformerDragger abort on unrecognized picked part (issue #174)

### LINKED ISSUE #174: Removal of "found" Assertion in SoTransformerDragger::dragStart()
**[Original report](https://bitbucket.org/Coin3D/coin/issue/174) by Anonymous.**

When working with multiple SoSelections, Manips and Draggers it can be quite frustrating when everything crashes because of this failed assertion. In my opinion an warning & continue approach would be much more appropriate. Thank you
```
#!shell

SoTransformerDragger.cpp:1187: void SoTransformerDragger::dragStart(): Assertion `found' failed.
Aborted
```

Dikluwe:
## Summary

Fixes [issue #174](https://github.com/coin3d/coin/issues/174): `SoTransformerDragger::dragStart()` could abort the application in a debug build (via `assert(found)`) when the pick path did not match any of its 20 recognized interactive part names.

This scenario arises in multi-`SoSelection`/multi-dragger setups and when the dragger is wired as a surrogate under an unrecognized part name via the documented `SoInteractionKit::setPartAsPath()` API.

### 1. Don't abort on unrecognized picked part (`b2d9f460a3`)
Replaces `assert(found)` in `dragStart()` with an early-return guarded by `COIN_DEBUG` warning. Also converts the `assert(0)` default branches in `drag()` and `dragFinish()` to no-ops for `WHATKIND_NONE`, so all three callbacks are safe throughout the full mouse button lifecycle.

### 2. Add end-to-end reproducer (`39ef47aced`)
`repro_surrogate.cpp` drives the real `SoDragger::handleEvent()` pick machinery via `SoHandleEventAction` (pure ray picking, no GL context) against a cube registered as a surrogate for `"surroundScale"` — a real catalog part whose name dragStart() does not recognize. Exercises mouse-down / move / up without abort.

### 3. Guard modifier events after ignored drag start (`42d3eede92`)
`SoTransformerDragger::SoTransformerDragger()`: initialize `ctrlDown` and `shiftDown` to `FALSE` (were uninitialized).  
`metaKeyChangeCB()`: add early-return for `whatkind == WHATKIND_NONE` to prevent the spurious `drag()` call that would otherwise occur while `isActive == TRUE` after an ignored `dragStart()`.

### 4. Add repro_metakey: targeted test for metaKeyChangeCB guard (`57be562d36`)
Closes the test gap: drives real `handleEvent()` modifier key events (LEFT_SHIFT, LEFT_CONTROL) while `isActive == TRUE` and `whatkind == NONE`, asserting no crash, `getCurrentState() == INACTIVE`, unchanged `getMotionMatrix()`, and zero `valueChanged` callbacks. Honestly documents the structural reason why reverting only the guard is not directly observable via the existing API (drag(WHATKIND_NONE) is already a no-op), while explaining why the guard is still correct and necessary. `run.sh` now builds and runs all three reproducers.

## Verification

- GCC and Clang (Debug), clean rebuild: all three reproducers pass.
- `ctest` (1/1): passes.
- Clang + ASan/UBSan on full CoinTests (330 tests, 83677 checks): zero new findings (3 pre-existing UB findings in unrelated code surfaced when running repro_surrogate under sanitizer directly — SoPath downcast pattern and SoCallbackList function-pointer-type mismatch — documented in the commit message; both are load-bearing pre-existing issues outside this fix's scope).
- MSan: blocked by a pre-existing scxml finding in `SoDB::init()`; the `ctrlDown`/`shiftDown` uninitialized-read component is documented in the reproducer header.

### DIFF
--- a/src/draggers/SoTransformerDragger.cpp
+++ b/src/draggers/SoTransformerDragger.cpp
@@ -811,6 +811,8 @@ SoTransformerDragger::SoTransformerDragger(void)
 
   this->state = INACTIVE;
   PRIVATE(this)->constraintState = CONSTRAINT_OFF;
+  PRIVATE(this)->ctrlDown = FALSE;
+  PRIVATE(this)->shiftDown = FALSE;
   // FIXME: according to SGI classdoc, this flag is supposed to be
   // default TRUE?  Investigate. 20011208 mortene.
   PRIVATE(this)->locateHighlighting = FALSE;
@@ -1099,7 +1101,8 @@ void
 SoTransformerDragger::metaKeyChangeCB(void *, SoDragger *d)
 {
   SoTransformerDragger * thisp = THISP(d);
-  if (!thisp->isActive.getValue()) return;
+  if (!thisp->isActive.getValue() ||
+      PRIVATE(thisp)->whatkind == WHATKIND_NONE) return;
 
   const SoEvent *event = thisp->getEvent();
   if (PRIVATE(thisp)->shiftDown != event->wasShiftDown()) {
--- a/src/draggers/SoTransformerDragger.cpp
+++ b/src/draggers/SoTransformerDragger.cpp
@@ -107,6 +107,7 @@
 #include <Inventor/actions/SoSearchAction.h>
 #include <Inventor/actions/SoGetMatrixAction.h>
 #include <Inventor/lists/SoPathList.h>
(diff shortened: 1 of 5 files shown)


Dikluwe:
Fechando: PR criada sem autorização do autor da branch. Será reaberta quando apropriado.

PR #726, opened 2026-09-13
Harden cc_dict resizing and hash changes

Dikluwe:
## Summary

- recycle superseded entries during resize instead of leaving them allocated in the dictionary allocator
- reindex existing entries when changing the hash function
- require a non-null hash callback, documented as a caller precondition and checked in Debug
- avoid an undefined empty-table statistic
- add regression coverage for collisions, resizing, clearing, randomized operations, hash changes, and boundary keys

`cc_dict` is used by `cc_heap`, scheduler/thread infrastructure, font and glyph caches, and OpenGL glue.

## Regression coverage

Coverage includes zero and pointer-shaped keys, null values, overwrite boundaries, complete collision chains, resize preservation, clearing and reuse, apply traversal, hash changes on populated and empty dictionaries, finite load-factor boundaries, empty statistics, and deterministic comparison against an independent model. A null hash callback is a caller precondition violation rather than a supported runtime configuration.

Changing the hash function after insertion lost existing mappings before this fix because entries remained linked under indices produced by the old function.

## Test plan

- GCC Debug full build and `ctest`: 8/8 passed
- GCC Release full build and `ctest`: 8/8 passed
- clang Debug full build and `ctest`: 8/8 passed
- GCC Debug with library and tests instrumented using `-fsanitize=address,undefined -fno-omit-frame-pointer`: 8/8 passed
- `git diff --check`: passed
- searched dictionary consumers and contract descriptions for stale hash-change assumptions

LeakSanitizer was disabled (`ASAN_OPTIONS=detect_leaks=0`) because the test environment runs under ptrace. Existing clang warnings in `SbVec*i32` are unrelated to this change.


Lgt2x:
@VolkerEnderlein @Dikluwe What's the point of these PRs? Do they solve an actual bug? These look very much like LLM slop to me, and I doubt they have been actually been tested against FreeCAD, which is the main consumer of Coin nowadays

VolkerEnderlein:
@Lgt2x I carefully evaluated the AI generated PRs and they address real issues and mostly W4 related warnings. I will not merge any PR that I feel might break Coin in any way.

Lgt2x:
Sounds good, thanks for your work!

Dikluwe:
Leaving some test results here for the record. They indicate that the resize change addresses a real internal memory-retention issue, although it is not a data-corruption bug.

`cc_dict_entry` objects are allocated from blocks owned by `cc_memalloc`. During a resize, the previous implementation allocated a replacement entry for every existing mapping and freed the old bucket array, but it did not return the old entries to `cc_memalloc`.

Consequently, those old entries remained physically allocated inside the allocator's memory blocks while being neither reachable through the new dictionary buckets nor present in the allocator's free list. They could therefore not be reused by subsequent insertions. Their memory was only released when the underlying `cc_memalloc` was cleared or destroyed.

I performed two temporary white-box experiments to verify this:

- With 8,192 live dictionary entries, the old implementation reported 16,412 active allocator units. After `cc_dict_clear()`, 8,220 units still remained active. With the fix, the counts were exactly 8,192 while populated and zero after clearing.

- In an address-tracing experiment, I recorded 98 existing entry addresses immediately before a resize. With the fix, all 98 addresses were reused during reindexing. With the old implementation, none were reused or added to the allocator's free list: all 98 became stranded inside the pool.

The tracing instrumentation was intentionally not committed because it required access to the private allocator layout. It was used only to independently confirm the behavior.

The test suite included in the PR covers:

- preservation of all mappings across repeated resizes;
- insertion, replacement, lookup, and removal;
- complete hash collisions and removal from every chain position;
- clear-and-reuse cycles;
- randomized operations checked against an independent model;
- hash-function replacement with reindexing of existing entries;
- resize-threshold and overwrite behavior;
- empty dictionaries, null values, zero keys, and pointer-shaped keys;
- load-factor boundaries;
- dictionary traversal and empty statistics.

The tests passed in the GitHub CI configurations for Ubuntu, Windows, and macOS, including static/shared and legacy variants where applicable.

I did not run a FreeCAD integration test, so I am not claiming that this was observed as a specific FreeCAD failure. The concrete impact is avoidable memory retention and allocator growth in long-lived dictionaries that resize.

PR #737, opened 2026-09-15
Clear the SoInput real-number parsing buffer

Dikluwe:
## Summary

- reset the reusable `SbString` numeric parsing buffer at the start of `readReal()`
- prevent characters retained from a previous numeric token from participating in the next real-number conversion
- use `makeEmpty(FALSE)` consistently with the integer readers while retaining the allocated buffer

## Integration

The branch now includes current `master`, including the merged `SbString` migration from #736.

## Validation

- clean Debug build with `COIN_BUILD_TESTS=ON`
- `CoinTests`: 342 tests and 83,783 checks passed
- CTest: 9/9 passed
- `git diff --check`: clean


src/io/SoInput_FileInfo.cpp
```diff
@@ -640,6 +640,8 @@ SoInput_FileInfo::readReal(double & d)
   int i, n;
   char c;
 
+  readString = "";
```
veelo: Now that #736 is merged, do you want to change this to `readString.makeEmpty(FALSE);` for consistency?

Dikluwe: Done in `d1d52bfa22` after merging the current master. `readReal()` now uses `readString.makeEmpty(FALSE)`, consistently with the other numeric readers. Full Debug build passed, `CoinTests` passed 342 tests and 83,783 checks, and CTest passed 9/9.

PR #754, opened 2026-09-18
Relink SbHash entries during resize

Dikluwe:
## Summary

- allocate the replacement bucket vector before mutating the existing table
- resize by relinking existing nodes instead of copying and destroying keys and values
- preserve key/value addresses across resize and avoid copy-constructor failures for existing mappings
- test the real fourth-insertion resize boundary, bucket growth, address stability, and copy/destruction counts

## Stack

Depends on #753 because relinking recalculates bucket indices under the non-throwing hash contract. Until the preceding PRs are merged, GitHub shows their commits here as well; the new change is the final relink-resize commit.

## Testing

- full six-commit stack: `ctest --test-dir /tmp/coin-sbhash-pr-build --output-on-failure` (8/8 passed)
- integration lab: 25/25 tests passed

This is PR 5 of 6 in the `SbHash` cleanup series.


Dikluwe:
Addressed the failing checks:

- the Windows CoinTests linker no longer calls the private, non-exported `SbHashFunc(const SbString&)` symbol directly; a Windows shared-build test support translation unit now supplies the test-only definition;
- the ABI workflow now suppresses only `SbHash` instantiations, which are explicitly documented as private implementation details.

Local validation: full shared Release build and 8/8 tests passed. I also reproduced the CI ABI comparison against v4.0.10 with RelWithDebInfo and installed headers; `abidiff` now reports only compatible additions (exit 4), with no removals.

PR #755, opened 2026-09-18
Defer SbHash storage allocation

Dikluwe:
## Summary

- defer bucket-vector and entry-pool allocation until the first `put()`
- keep `clear()` optimized for reuse and add `releaseStorage()` for callers that need to return retained storage
- make empty lookup, erase, iteration, copying, statistics, and destruction safe before allocation
- preserve the historical pool-before-buckets release order to avoid the teardown regression found by the local benchmark

## Measurements

- untouched default maps: approximately 2,160 to 48 heap bytes per map in the local allocation probe
- empty construction/destruction: approximately 98% faster
- first insertion with the default size: approximately 18% faster
- no stable regression remained in growth, lookup, iteration, or repeated `put`/`erase` after restoring release order

The benchmark is intentionally kept outside this PR and is used only as local engineering evidence.

## Stack

Depends on #754. Until the preceding PRs are merged, GitHub shows their commits here as well; the new change is the final lazy-storage commit.

## Testing

- isolated Coin build: 8/8 tests passed
- integration lab: 25/25 tests passed
- FreeCAD embedded Coin and `pivy_coin` rebuilt successfully
- FreeCAD headless smoke test: `SBHASH_FREECAD_OK Separator`

This is PR 6 of 6 in the `SbHash` cleanup series.


Dikluwe:
Rebased this stacked PR onto the corrected #754 and addressed its remaining ABI report. Lazy storage stops emitting two constructor symbols for `SoGLDriverDatabaseP::SoGLDriver`, a type defined only in `SoGLDriverDatabase.cpp`. The suppression matches exactly those C1/C2 mangled symbols instead of hiding the class broadly.

Local validation: full shared Release build and 8/8 tests passed. The exact CI ABI comparison against v4.0.10 now reports only compatible additions (exit 4); the two private symbols are filtered and no ABI removals remain.

PR #756, opened 2026-09-18
Make SbList growth exception safe

Dikluwe:
## Summary

- keep the candidate capacity local until allocation and value transfer succeed
- keep `<memory>` out of the widely included public header and clean up candidate arrays explicitly
- reject signed capacity overflow with `std::bad_alloc`
- preserve the original list if element assignment throws in either `grow()` or `fit()`
- cover allocation, construction, and assignment failure recovery

## Problem

`SbList::grow()` updated `itembuffersize` before `new Type[]`. If allocation or array construction threw, the old four-element buffer remained active while the recorded capacity became eight. A later append could then write beyond the inline buffer.

`fit()` also allocated a candidate heap buffer without releasing it if element assignment threw. Both operations now commit the pointer and capacity only after value transfer succeeds.

This remains separate from #738, which addresses removal of an absent item. The current branch includes the merged #738 through `master`.

## Review follow-up

- replaced `std::unique_ptr` with a local raw candidate plus `try`/`catch`, avoiding `<memory>` in every client translation unit
- updated the obsolete growth comment
- added failure-injection tests for assignment during both `grow()` and `fit()`

## Validation

- clean Debug build with `COIN_BUILD_TESTS=ON`
- `CoinTests`: 345 tests and 83,826 checks passed
- CTest: 9/9 passed
- `git diff --check`: clean

The contract commit still precedes the implementation and review-fix commits in the branch history.


include/Inventor/lists/SbList.h
veelo: `DEFAULTSIZE` no longer applies. Please rephrase to match the code.

include/Inventor/lists/SbList.h
```diff
@@ -34,7 +34,10 @@
 \**************************************************************************/
 
 #include <cassert>
+#include <climits>
 #include <cstddef> // NULL definition
+#include <memory>
```
veelo: `SbList.h` is included all over the API. `std::unique_ptr` is only needed for the few lines of `grow()`. You can get the same unwind safety with a local raw pointer and a try/catch, or a tiny local guard, and keep `<memory>` out of every client TU.

veelo (changes requested):
Review courtesy Grok. It notes:

> `fit()` still does `new Type[items]` without a guard and can leak the new buffer if a later assignment throws. Out of scope if you only care about the ASan overflow.

include/Inventor/lists/SbList.h
Dikluwe: Addressed in `537b690ce5`. `<memory>` is no longer included by the public header. `grow()` uses a local raw candidate pointer with `try`/`catch`, deletes it if element assignment throws, and commits the pointer and capacity only after the copy succeeds.

include/Inventor/lists/SbList.h
Dikluwe: Rephrased in `537b690ce5`. The comment now describes the actual default operation and the signed-overflow guard, without referring to the removed `DEFAULTSIZE` fallback.

Dikluwe:
The general review note about `fit()` is addressed as well in `537b690ce5`: a heap candidate is now released if element assignment throws, while the original buffer, capacity, length, and values remain active. Added separate failure-injection coverage for assignment failures in both `grow()` and `fit()`. Full Debug build passed, `CoinTests` passed 345 tests and 83,826 checks, and CTest passed 9/9.

veelo:
Great! Many thanks.

